PR #41, opened 2016-03-23
Adding transform CONSECTIVEABOVE

aertoria:
Adding transform CONSECTIVEABOVE.

Given a time series, capture the data points meets the requirement below, to form a new time series.
This transform captures all the data points whose value is above threshold1, and continuesely for threshold2 consective data points.

ethan.wang@salesforce.com

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ logs/
 nbactions.xml
 nb-configuration.xml
 argus-build.properties
+.DS_Store
--- a/ArgusCore/src/test/java/com/salesforce/dva/argus/AbstractTest.java
+++ b/ArgusCore/src/test/java/com/salesforce/dva/argus/AbstractTest.java
@@ -1,4 +1,4 @@
-/*
+ /*
  * Copyright (c) 2016, Salesforce.com, Inc.
  * All rights reserved.
  *
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/ConsectiveAboveValueMapping.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/ConsectiveAboveValueMapping.java
@@ -96,36 +96,4 @@ public class ConsectiveAboveValueMapping implements ValueMapping{
 		return TransformFactory.Function.CONSECTIVEABOVE.name();
 	}
 }
-
-
-//Another implementation. Much more efficent but this stacks up in memory, so please don't use this one.
-//private Map<Long, String> originalDatapoints;
-//private ArrayList<Long> keyList;
-//private Map<Long, String> resultMetric = new TreeMap<Long, String>();
-//private int threshold=0;
-//private int consective=0;
-//
-//private Object crawler(int i,int cum){
-//	if(i==keyList.size()){	
-//		if(cum>=consective){
-//			for(long j:keyList.subList(i-cum, i)){
-//				System.out.println(j);
-//			}
-//		}
-//		return null;
-//	}
-//	Double value=Double.parseDouble(originalDatapoints.get(keyList.get(i)));
-//	System.out.println("currently-"+value);
-//	if (value>=threshold){//collect you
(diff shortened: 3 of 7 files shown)


.gitignore
```diff
@@ -9,3 +9,4 @@ logs/
 nbactions.xml
 nb-configuration.xml
 argus-build.properties
+.DS_Store
```
saaja-sfdc: DS_Store should not be committed.

aertoria: That's why I add it to .gitignore. so that all .DS_Store will not be commited

ArgusCore/src/test/java/com/salesforce/dva/argus/AbstractTest.java
```diff
@@ -1,4 +1,4 @@
-/*
+ /*
```
saaja-sfdc: Original formatting has to be preserved in the header.

saaja-sfdc:
@aertoria This transform performs two operations, one of which already exists (CULL_ABOVE/CULL_BELOW).  Can we write just CONSECUTIVE_ABOVE, CONSECUTIVE_BELOW transforms and then you can achieve the functionality by joining the results of a CULL_BELOW and CONSECUTIVE_BELOW transform on a single expression.  This would be generally more useful than a single transform that combines both operations.

aertoria:
I like that idea. In that case I will only have one transform called CONSECUTIVE. Following after either cull_above or cull_below, CONSECUTIVE will filter away any data points that is not consecutive based on threshold.

saaja-sfdc:
@aertoria Then I will close this PR.  You can implement a CONSECUTIVE transform class but please allow the choice of filtering above and below the threshold and expose it in the grammar as CONSECUTIVE_ABOVE & CONSECUTIVE_BELOW so it's consistent with other filter transforms.

aertoria:
@saaja-sfdc I'm not sure I'm following your last comments completely. You mentioned 
> to allow CONSECUTIVE the choice of filtering above and below the threshold and expose it in the grammar as CONSECUTIVE_ABOVE & CONSECUTIVE_BELOW

I was thinking to write CONSECUTIVE as one uni-sex mapping. Its only function is to take in a series, find out all consecutive data points and toss them into an new series and output it. 
In that way, if user want to do CONSECUTIVE_ABOVE to get all the data points in a series whose value is above 100 for at least consecutive 5 data points*, user will use it in this way:

CONSECUTIVE(CULL_BELOW(-1h:scope:mem.heap.used:avg,$100,$value), $5)

Does this kind of align with what you envisioned?

saaja-sfdc:
@aertoria I understand.  So please disregard my previous statement; I think just a single CONSECUTIVE transform is fine.  However, since Argus doesn't require a regular interval for datapoints, you might consider changing the parameter from a scalar value to a time interval and updating the semantics to:

CONSECUTIVE: Returns all datapoint's for which the interval between adjacent datapoints is less than or equal to the specified time interval.

e.g. CONSECUTIVE(CULL_BELOW(-1h:scope:mem.heap.used:avg,$100,$value), $5m)

aertoria:
@saaja-sfdc  Thanks for clarification. Regarding that $5m, actually I had a thought about that too. It's actually a bit complicated. That's why in the last comments I didn't bring up, I put an asterisk instead.

> Returns all datapoint's for which the interval between adjacent datapoints is less than or equal to the specified time interval.
Basically I like this design because it's intuitive and highly reusable for general purposes. My only concern was, from me and Tony perspective, this might not be able to tell if the datapoint are _seamlessly_ adjacent. 

Consider a scenario that during 5 min we have 5 data points, with only one of them in the middle that is below threshold. Since CONSECUTIVE will only start to judge _after_ CULL_BELOW, that bad data points will firstly get throw away. As a result the end result will looks like adjacent. I don't know if that make sense.

So basically I was thinking something like this
CONSECUTIVE(series, $5m,$1m)
The first $5m is the same, the second $1m tells the default resolution.

PR #63, opened 2016-06-14
Adding DOWNSAMPLE transform by count feature

aertoria:
Adding DOWNSAMPLE transform by count feature
@saaja-sfdc @rajsarkapally-sfdc 

Discussed with Raj, this is a very useful feature to have. It will downsample a time series based on the count of how many data points in each time window.


saaja-sfdc:
Looks good @aertoria

aertoria:
@saaja-sfdc Thanks!

PR #66, opened 2016-06-15
Distributed scheduling using database

rajsarkapally-sfdc:
@saaja-sfdc  merged with the latest code.


ArgusCore/src/main/java/com/salesforce/dva/argus/service/ServiceFactory.java
```diff
@@ -85,6 +85,8 @@
     Provider<CacheService> _cacheServiceProvider;
     @Inject
     Provider<DiscoveryService> _discoveryServiceProvider;
+    @Inject
+    Provider<DistributedSchedulingLockService> _distributedSchedulingLockServiceProvider;
```
saaja-sfdc: @rajsarkapally-sfdc  This can be removed from the ServiceFactory.  The lock service provider is only used by the distributed scheduling service and shouldn't be used generally by other code.  I think it's enough to just add the binding to SystemInitializer and inject an instance into the constructor of DefaultDistributedSchedulingService.  You're doing both of these already, so you only need to remove this code from ServiceFactory.

rajsarkapally-sfdc: @saaja-sfdc  Instance of DefaultDistributedSchedulingService is required for writing JUnit test cases similar to GlobalInterLock. 

DistributedSchedulingLockServiceTest requires instance of DefaultDistributedSchedulingService.

If I instantiate with constructor, Guice will not inject objects like entityManager.

saaja-sfdc: You can still use DI.  Just instantiate in your unit test using the 'new' keyword and populate the ctor parameters with services obtained from the ServiceFactory.  This shouldn't be exposed in the service factory.  GlobalInterlock is a general mechanism, but this is only used for disto scheduling and moreover, just for this specific implementation.

saaja-sfdc:
@rajsarkapally-sfdc This looks good, but tests are failing.  Can you please review and get the tests squared away so we can merge?

rajsarkapally-sfdc:
@saaja-sfdc  Fixed unit test cases.

PR #67, opened 2016-06-17
1)Return admin user instead of default user in NoAuthService. 2)Corre…

dilipdevaraj-sfdc:
…ct endpoint for wardenpolicyforuser


saaja-sfdc:
@dilipdevaraj-sfdc @kgowdru Can this be changed to do two things:  1) create a user for whatever user name is passed in with the credentials and 2) have a NoAuthService configuration property to indicate whether all authentication returns privileged users or not (service.auth.noauthservice.privileged=[true | false]).

dilipdevaraj-sfdc:
@saaja-sfdc 
Updated with changes.Please review.

PR #71, opened 2016-06-22
Minor UI updates to code base, templates, headings, etc.

cnardi-dev:
Also added a new directive to autofocus to input fields, such as the Login view, to benefit the accessibility and usability of the username field, and allow the user to easily start typing without having to use the mouse to focus the field before typing.

Other updates include minor tweaks to html & css settings to normalize the layout, headings, spacing, alignment, etc.  More updates of this type will be coming in the future, including splitting up headers, tabs, etc. into separate directives to help keep templates more consistent across the site.


saaja-sfdc:
Looks good @cnardi-dev

PR #75, opened 2016-07-06
Add in full join feature for handing situations when two data points do not share common time stamp.

aertoria:
This feature added in at  MetricReducerOrMappingTransform.java, which will benefit four transform:
SUM
DIVIDE
DIFF
SCALE.

Example:
Before this feature:    (1L:1  2L:2)    +  (2L:2,3L:3)  = (2L:4) 
After this feature:    (1L:1  2L:2)    +  (2L:2,3L:3)  = (1L:1, 2L:4, 3L:3)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ nbactions.xml
 nb-configuration.xml
 argus-build.properties
 .DS_Store
+.idea
+.Project
+.classpath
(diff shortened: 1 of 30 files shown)


saaja-sfdc:
@aertoria Can you rename the constant from JOIN to UNION?  That's the commonly used term for this type of operation.  JOIN by itself is infers an inner join which is an intersection operation.

aertoria:
Thanks for comment. Closing this pr. I open another one

PR #76, opened 2016-07-07
add feature for handing the operations when two data points do not share common time stamp

aertoria:
Changing from 'full' to 'union' as optional constant

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ nbactions.xml
 nb-configuration.xml
 argus-build.properties
 .DS_Store
+.idea
+.Project
+.classpath
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ argus-build.properties
 .idea
 .Project
 .classpath
+property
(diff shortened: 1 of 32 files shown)


saaja-sfdc:
@aertoria Can you please rebase so I can merge.  Before I actually merge it, I'll request that you update the [wiki](https://github.com/SalesforceEng/Argus/wiki/Transforms) with the info for the new parameter.  Once the documentation is updated, I'll go ahead and merge the change.

Just something to the effect that by default, it does an intersection across timestamps and that you can specify a union with the parameter.  It does need to be reflected in the wiki for any affected transform.

Thanks for contributing!

aertoria:
@saaja-sfdc  Thanks for working on this.
1.Let me know if you are able to merge now, or you still want me to do a REBASE.
2.What's the best practice of contributing content on wiki?  I have wiki clone locally and updated, But I don't have permission to directly write back. Plz advise

saaja-sfdc:
Hi @aertoria , yep we are working on getting contribution guidelines in place.  For wiki changes, you should fork the wiki and then open an issue as described in this process.

https://gist.github.com/larrybotha/10650410

aertoria:
@saaja-sfdc 
Thanks Tom! 
Let me know when you done this pr. some other stuff in my other branches will follow.

PR #82, opened 2016-07-27
STL Anomaly Detection

muruvig:
Added a transform called by ANOMALY_STL(<timeseries>) and ANOMALY_STL(<timeseries>, $constant). The constant in the second call gives the user the option to select the size of a season. Transform currently returns the anomaly score between 0 and 100.


saaja-sfdc:
@muruvig It looks like this failed to compile.  Perhaps there's a file missing from the PR?

saaja-sfdc:
@muruvig I merged this from my develop branch.  I updated the wiki with basic usage, but you might want to update it to provide sample input/output as other transforms do.

PR #83, opened 2016-07-28
Update alert expression max length to 2048.

saaja-sfdc:


bsura:
@saaja-sfdc Looks good. Although the branch is out-of-date. Can you rebase so that we can merge it with the base branch?

PR #84, opened 2016-08-01
Metric Search and Discovery UI

shouvikmani:
The Argus Metric Search and Discovery UI will enable users to identify what metrics exist in Argus, allowing them to query for metrics and visualize metric hierarchies. There will also be improvements to the discovery web service endpoint to support the UI functionality.


cnardi-dev:
Is it possible to change or re-submit this PR to the 'develop' branch instead?

PR #96, opened 2016-08-12
Base D3 line chart directive

colbyguan:
Should work for multiple lines. Changed mouseover tooltip to not follow the mouse but rather stay in the same place. 

Example usage is in Metrics tab, select "Use D3"

### DIFF
--- a/ArgusWeb/app/js/config.js
+++ b/ArgusWeb/app/js/config.js
@@ -20,7 +20,7 @@
 angular.module("argus.urlConfig", [])
 .constant('CONFIG', {
     version: '4.0-SNAPSHOT',
-    wsUrl: 'http://localhost:8080/argusws/',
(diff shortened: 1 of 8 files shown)


ArgusWeb/app/js/config.js
```diff
@@ -20,7 +20,7 @@
 angular.module("argus.urlConfig", [])
 .constant('CONFIG', {
     version: '4.0-SNAPSHOT',
-    wsUrl: 'http://localhost:8080/argusws/',
+    wsUrl: 'http://localhost:8080/argus/',
```
colbyguan: Should we just gitignore this file since it is created by grunt?

cnardi-dev: Yes, no need to commit this.

cnardi-dev:
Looks good! Thnx for creating this D3 directive.

PR #100, opened 2016-08-17
hot fix to multi-select for assigning triggers to notifications.

cnardi-dev:
short term fix to help users identify triggers to assign to a notification.


cnardi-dev:
Thanks Dilip!

PR #109, opened 2016-08-24
Parametrized Url updates

cnardi-dev:
- automatically add control params to url when dashboard loads.
- allow the user to update the local scope of controls, and also update params in the url.


cnardi-dev:
What can I do to fix this failed check?

PR #123, opened 2016-08-30
Add tracking of daily and monthly users.

saaja-sfdc:
Updated to track mau/dau based on duration of entries.
@SalesforceEng/argus please review

### DIFF
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/auth/LDAPAuthService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/auth/LDAPAuthService.java
@@ -77,6 +77,8 @@ public class LDAPAuthService extends DefaultService implements AuthService {
     private final UserService _userService;
     private final SystemConfiguration _config;
     private final MonitorService _monitorService;
+    private final UserCountCache _monthlyUsers;
+    private final UserCountCache _dailyUsers;
 
     //~ Constructors *********************************************************************************************************************************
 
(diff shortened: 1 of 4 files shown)


ArgusCore/src/main/java/com/salesforce/dva/argus/service/auth/LDAPAuthService.java
```diff
@@ -196,7 +200,11 @@ private PrincipalUser _findPrincipalUser(String userDn, String username, String
                 result = new PrincipalUser(_userService.findAdminUser(), username, email);
                 result = _userService.updateUser(result);
             }
-            _monitorService.updateCounter(MonitorService.Counter.UNIQUE_USERS, _userService.getUniqueUserCount(), new HashMap<String, String>(0));
+            _dailyUsers.put(username, System.currentTimeMillis());
+            _monthlyUsers.put(username, System.currentTimeMillis());
+            _monitorService.updateCounter(MonitorService.Counter.DAILY_USERS, _dailyUsers.size(), new HashMap<>(0));
+            _monitorService.updateCounter(MonitorService.Counter.MONTHLY_USERS, _monthlyUsers.size(), new HashMap<>(0));
```
dilipdevaraj-sfdc: We could leverage this similar to how we get unique user count from database over time. This si written through multiple web servers hence we would need to use 
expression=-1d:argus.core:users.unique:avg

saaja-sfdc: @dilipdevaraj-sfdc another option is to update the lastModifiedDate on userPrincipal object and do a select count(id) where lastModifiedDate > a day ago. We could also add a lastLoginDate field to principalUser field to facilitate this. That way we get it straight from the db and it would be accurate across all instances. Thoughts?

dilipdevaraj-sfdc: Yes we could do it this way  by reading from database.
But by having a time series for each web server (the way you implemented it) we can also count the login count load distribution across webservers.

PR #143, opened 2016-09-10
fix agDate

prestonfff:


ArgusWeb/app/js/directives/controls/date.js
```diff
@@ -8,6 +8,15 @@ angular.module('argus.directives.controls.date', [])
             controlValue: '@default'
         },
         controller: function($scope, $filter) {
+            $scope.ctrlVal = $scope.controlValue;
+
+            // check $routeParams to override controlValue
+            for (var prop in $routeParams) {
```
cnardi-dev: We should use the new Controls service for the $routeParams comparison.

PR #144, opened 2016-09-12
1) Write alerts scheduled logs as a time series to TSDB 2) Limit max …

dilip-devaraj:
…number of alerts per scheduler in distributed database scheduling so distribution is not skewed

### DIFF
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
@@ -52,6 +52,7 @@ import com.salesforce.dva.argus.service.MetricService;
 import com.salesforce.dva.argus.service.MonitorService;
 import com.salesforce.dva.argus.service.MonitorService.Counter;
 import com.salesforce.dva.argus.service.NotifierFactory;
+import com.salesforce.dva.argus.service.TSDBService;
 import com.salesforce.dva.argus.service.jpa.DefaultJPAService;
 import com.salesforce.dva.argus.service.metric.transform.MissingDataException;
 import com.salesforce.dva.argus.system.SystemConfiguration;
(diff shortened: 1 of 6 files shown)


dilip-devaraj:
Adding  reference of AlertService using MonitorService. This option was not pursued since if we have a large numebr of alerts per web service machine then (eg > 20,000) then the monitoring service does not write its data consistently every minute. For a period of an hour instead of seeing 59/60 datapoints we see around 43-50 datapoints.
https://gist.github.com/dilip-devaraj/559631fd48009139c7c4d51ddbe268c4

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
```diff
@@ -573,15 +578,23 @@ public void enqueueAlerts(List<Alert> alerts) {
...
 		for (Alert alert : alerts) {
-			if (findAlertByPrimaryKey(alert.getId()) == null) {
-				_logger.warn("Could not find alert ID {}", alert.getId());
-				continue;
-			}
-			_historyService.createHistory(addDateToMessage("Alert queued for evaluation"), alert, JobStatus.QUEUED, 0, 0);
+			Map<Long, String> datapoints = new HashMap<>();
+			// convert timestamp to nearest minute since cron is Least scale resolution of minute
+			datapoints.put(1000*60 * (System.currentTimeMillis()/(1000 *60)), "1");
+			Metric metric = new Metric("alerts.scheduled", "alert-" + alert.getId().toString());
```
bsura: I would suggest using metric name as "<alertID>-alert". This would improve retrieval of data for that alert.

dilipdevaraj-sfdc: As discussed since we are retrieving the exact record during retrieval the performance will be same if we alert-1 or 1-alert. So not changing this.

PR #170, opened 2016-09-22
Fix series name & color for multiple ag-metric

cnardi-dev:
This PR also has a merge from the 'develop' branch in case we are merging from that... if not, can cancel this.


dilipdevaraj-sfdc:
@cnardi-dev  Can you please resolve conflicts and resubmit PR

PR #186, opened 2016-09-25
Update to correctly capture MAU/DAU across all instances of AuthService.

saaja-sfdc:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/auth/LDAPAuthService.java
```diff
@@ -200,10 +200,10 @@ private PrincipalUser _findPrincipalUser(String userDn, String username, String
                 result = new PrincipalUser(_userService.findAdminUser(), username, email);
                 result = _userService.updateUser(result);
             }
-            _dailyUsers.put(username, System.currentTimeMillis());
-            _monthlyUsers.put(username, System.currentTimeMillis());
-            _monitorService.updateCounter(MonitorService.Counter.DAILY_USERS, _dailyUsers.size(), new HashMap<>(0));
-            _monitorService.updateCounter(MonitorService.Counter.MONTHLY_USERS, _monthlyUsers.size(), new HashMap<>(0));
+            DAILY_USERS.put(username, System.currentTimeMillis());
+            MONTHLY_USERS.put(username, System.currentTimeMillis());
```
dilipdevaraj-sfdc: Agree, since a new object of LDAP is instantiated each time on login we can use the static map to maintain count

We can then do a user count by doing a transform on -1d:argus.core:users.daily:sum

PR #207, opened 2016-10-08
fix issue of date picker only pick local time. 

aertoria:
fix issue of date picker only pick local time.
now it is returning GMT time, which align up with argus.

For the date time that input from URL, or as format as '-5d', it will not convert.


aertoria:
@cnardi-dev @saaja-sfdc

PR #212, opened 2016-10-11
In monitoring service catch Throwable, so we can catch any Errors, in…

dilipdevaraj-sfdc:
… addition to Exception


dilipdevaraj-sfdc:
@bsura Can you please review.

PR #221, opened 2016-10-17
Add new time toggle buttons

prestonfff:
-Add new time toggle buttons
-refine some codes on brush&zoom
-fix the brush&zoom/pan not syncing correctly bug


prestonfff:
Crossline and Tooltipbox missing. Will fix that in next PR

PR #223, opened 2016-10-18
Get alerts in batches and add trigger fired timestamp to notifier.

bsura:


bsura:
@saaja-sfdc @rajsarkapally-sfdc

PR #230, opened 2016-10-21
Add snapping to fill transform when it generates a constant line

xizi-xu:
I modified test cases as well.


xizi-xu:
To test this feature:

1. Use an expression to generate a constant line. Example: FILL(#-5h#, #-0h#, #1h#, #0m#, #100#).
2. The graph's timestamps should be at a uniform boundary based on the interval provided and then the offset is applied. If the above expression is run, the timestamps will be rounded to the exact hour. 
3. If the interval is greater than the different between start and end point, there should be only 1 data point in the graph.

PR #236, opened 2016-10-25
1)Make RedisCacheService singleton so same JedisCluster object is reu…

dilipdevaraj-sfdc:
…sed for performance optimization 2) Add log in QuartzScheduling


dilipdevaraj-sfdc:
@bsura  Can you review ?

PR #238, opened 2016-10-25
chart updates: create new service to render line chart. remove calls …

cnardi-dev:
…to Highchart rendering. add some container styling and resize element proportions. add html template for the top toolbar. move legend to top of graph.

need to fix tabbing as well :(


xizi-xu:
When you view a particular dashboard and submit a new metric, the old graph stays and the new graph is added at the bottom without the actual graph
screen shot 2016-10-25 at 2 26 37 pm

cnardi-dev:
Thanks John, I will clear out the container before another redraw.

PR #239, opened 2016-10-25
Log the exception when failing to send an email

bsura:


bsura:
@rajsarkapally-sfdc Please review and merge.

PR #245, opened 2016-10-27
Alert Loading Enhancement

xizi-xu:
- New GET API call for the alert list (/alerts/meta). It returns all alerts with empty expressions, triggers, and notifications.

- SessionStorage for the alert list. The reload button, editing an alert in its particular page with saving, and open a new browser tab can trigger a new API call for alert list.

### DIFF
--- a/ArgusWeb/app/js/config.js
+++ b/ArgusWeb/app/js/config.js
@@ -1,8 +1,8 @@
 /*! Copyright (c) 2016, Salesforce.com, Inc.
  * All rights reserved.
- *
+ *  
  * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
- *
+ *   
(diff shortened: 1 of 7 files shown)


xizi-xu:
Before testing: rebuild ArgusWebService and ArgusWeb.

To test:

1. [http://localhost:8080/argusws/alerts/meta](url) is the API endpoint for alert list query with no content. All the alerts should have empty expressions, noftificationIds, and triggersIds. 
2. Switching between different tabs within Argus (such as Dashboards and Metrics) should not trigger a new API call over the network. However, the blue reload button on the top right of the table should.
3. Add, delete, enable, and edit an alert should all behave normally as usual. The list should update itself according.

bsura:
Oops. I had some comments for this one. Looks like @cnardi-dev merged it. I will discuss it with @xizi-xu . I am doubtful if this is gonna speed up the loading of alerts. Let's take it offline.

PR #246, opened 2016-10-28
GMT Option Bug Fix

xizi-xu:
- Disable GMT option when offset is used for agDate input.
screen shot 2016-10-28 at 8 15 08 am

- Update URL param such that " GMT" is added to URL properly when GMT is used for agDate. Existing links will work normally without GMT checkbox being checked. This does not apply to offset input for agDate
Existing link with no GMT:
screen shot 2016-10-28 at 8 18 53 am
New link with GMT:
screen shot 2016-10-28 at 8 19 44 am


xizi-xu:
To test: 
1. Switch between offset input and datepicker input for agDate to see if GMT checkbox is disabled and enabled.
2. Copy and paste the link of a dashboard with GMT checkbox checked in a new browser tab. The same dashboard will show up. Same thing with GMT checkbox unchecked.

PR #253, opened 2016-11-01
Make email exceptions while evaluating alerts to be configurable

dilipdevaraj-sfdc:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
```diff
@@ -539,7 +541,7 @@ private boolean shouldEvaluateNotification(Notification notification, BigInteger
 	private void sendEmailToAdmin(Alert alert, BigInteger alertId, Throwable ex) {
 		Set<String> to = new HashSet<>();
 
-		to.add(_configuration.getValue(com.salesforce.dva.argus.system.SystemConfiguration.Property.ADMIN_EMAIL));
```
xizi-xu: Why was this changed?

PR #254, opened 2016-11-01
Merge commits from release branch (2.2) into develop

bsura:
Created a new one since we were having problems merging the old one.


xizi-xu:
For some reason, I cannot update the 2.2 branch. The merge is not allowed until the update.

PR #258, opened 2016-11-02
Enable dashboard and alert lists reload with page refresh

xizi-xu:
With this change, dashboard and alert query calls can be trigger by refreshing the browser tab as well.


xizi-xu:
To test:
1. Open Argus in 2 separate browser tabs.
2. Create or delete dashboards in one of the tabs.
3. Refresh the other tab (command + r) and the dashboard list will be updated accordingly.

PR #261, opened 2016-11-03
Add time domain to x axis

xizi-xu:
X axis now can display local time or GMT based on user's input value in agDate.

### DIFF
--- a/ArgusWeb/app/js/directives/charts/chart.js
+++ b/ArgusWeb/app/js/directives/charts/chart.js
@@ -3,26 +3,27 @@ angular.module('argus.directives.charts.chart', [])
 function(Metrics, ChartRenderingService, ChartDataProcessingService, ChartOptionService, CONFIG, VIEWELEMENT, $compile) {
     var chartNameIndex = 1;
 
-    function renderLineChart(scope, newChartId, series, updatedAnnotationList) {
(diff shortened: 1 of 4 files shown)


ArgusWeb/app/js/directives/charts/chart.js
```diff
@@ -3,26 +3,27 @@ angular.module('argus.directives.charts.chart', [])
...
+        // assign chartId, series data, time domain to new $scope
         lineChartScope.chartId = newChartId;
         lineChartScope.series = series;
-
+        lineChartScope.startTime = startTime;
+        lineChartScope.endTime = endTime;
+        debugger;
         // append, compile, & attach new scope to line-chart directive
-        angular.element("#" + newChartId).append( $compile('<line-chart chartid="chartId" series="series"></line-chart>')(lineChartScope) );
+        angular.element("#" + newChartId).append( $compile('<line-chart chartid="chartId" series="series" starttime="startTime" endtime="endTime"></line-chart>')(lineChartScope) );
```
prestonfff: That is the current solution, which picks min and max from datapoints and set the x axis domain.

prestonfff: @cnardi-dev @xizi-xu We can determine the start/end and domain without passing the extra parameters from controller. The previous brush xAxis showing 2000&Jan 02 at the two ends was due to brush xAxis not being redrawn (even the domain was set).

PR #262, opened 2016-11-04
Resolve the problem of Loading spinner not showing up after clicking reload button 

xizi-xu:


cnardi-dev:
looks good. merging to also update the d3 branch.

xizi-xu:
Merged too early... found another one liner bug lol @cnardi-dev

PR #265, opened 2016-11-04
Minor changes and Fix mouse over line path zoom/pan/drag not working bug

prestonfff:


prestonfff:
@cnardi-dev  This is just fixing the previous bug that you can not drag when cursor sitting on the line when the lines getting crowded. The brush by dragging on the main chart is not implemented yet. Will work on that today.

PR #271, opened 2016-11-08
Minor fixes on d3 linechart

prestonfff:
- Fix an issue that the circle points on line do not sync when zooming/panning on the chart.
- Fix some problems caused by last changes.


cnardi-dev:
good fixes! merging.

PR #273, opened 2016-11-08
A few UI hot fixes (resubmission)

xizi-xu:
1. Allow browser refresh to trigger API query call for dashboard/alert lists.
2. Show loading spinner after the reload button is pressed.
3. Prevent tab switching within dashboard list when reload button is pressed.

All of the above are in the develop branch as well.


xizi-xu:
@cnardi-dev @bsura @saaja-sfdc can you guys re-review this?

PR #277, opened 2016-11-10
Use legend to toggle graph on and off

xizi-xu:
- Click graph name in the legend to toggle it on and off
- Hiding others feature is supported as well
- Once a graph is hidden, its name will turn gray in the legend (this is a little bit slow and laggy. Maybe @cnardi-dev can help?)


cnardi-dev:
Hard to pinpoint the lagginess without looking at it in the browser.  Is it still working off the mousemove event?  That would definitely make it laggy.

PR #279, opened 2016-11-12
render Annotation as flag. Merging from develop branch.

cnardi-dev:
see the TODOs in the comments.  know issues:

- x position of annotation is not syncing with x axis (also update on panning/zooming)
- need to calc offset to position for multiple annotations
- text field is too large and can't be used.  need alternate way to inspect/copy description

### DIFF
--- a/ArgusWeb/app/js/directives/charts/lineChart.js
+++ b/ArgusWeb/app/js/directives/charts/lineChart.js
@@ -15,7 +15,6 @@ angular.module('argus.directives.charts.lineChart', [])
         templateUrl: 'js/templates/charts/topToolbar.html',
         controller: ['$scope', function($scope) {
             $scope.toggleSource = function(source) {
-                debugger;
                 toggleGraphOnOff(source);
             };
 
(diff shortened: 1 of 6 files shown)


cnardi-dev:
screen shot 2016-11-11 at 6 29 42 pm

ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -589,6 +584,42 @@ angular.module('argus.directives.charts.lineChart', [])
...
+            function updateAnnotations() {
+                if (!scope || scope.series.length > 1 ) return;
+
+                var flagSeries = scope.series[0].flagSeries.data
+                var flagsG = d3.select('svg').select('.flags');
+                var label = flagsG.selectAll("flagItem")
+                    .data(flagSeries)
+                    .enter().append("g")
+                    .attr("class", "flagItem")
+                    .attr("transform", function(d) {
```
xizi-xu: This is just a hard code value right?

cnardi-dev: Not harding the data, its making an API call specifically for the Annotations, which adds 'flagSeries', to the 'series' object before the lineChart directive is called.

cnardi-dev: flagseries-data

cnardi-dev: If there is NO flagSeries data, then this object will not be present.

PR #280, opened 2016-11-12
Cross line tooltip and main chart brush feature

prestonfff:
- Fix some minor issues.
- Refine some codes
- Move the crosslineTooltip info to X & Y Axis and add some styles.
- Implement main chart brush and add checkbox for it.
- Fix the issue that y domain does not include some datapoints
- Fix an issue that full brush will show when resizing

### DIFF
--- a/ArgusWeb/app/js/directives/charts/lineChart.js
+++ b/ArgusWeb/app/js/directives/charts/lineChart.js
@@ -530,7 +530,6 @@ angular.module('argus.directives.charts.lineChart', [])
                 reScaleY(); //rescale domain of y axis
                 //redraw
                 redraw();
-
                 // sync the brush
                 context.select(".brush").call
                 (brush.move, x.range().map(t.invertX, t));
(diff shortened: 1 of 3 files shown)


ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -51,14 +51,28 @@ angular.module('argus.directives.charts.lineChart', [])
             // angular.element('').on('click', function() {
             //     scope.$apply();
             // });
+
             // toggle source to hide/show, leave other sources showing
+            scope.toggleSource = function(source) {
```
xizi-xu: These should not be re-commit back to the codebase; they were deleted from previous PR. Please pull the newest from upstream before pushing creating a PR.

ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -360,41 +391,65 @@ angular.module('argus.directives.charts.lineChart', [])
...
+                focus.select('[name=crossLineTipX]')
+
                     .attr('x', X)
                     .attr('y', height )
                     .attr('dy', crossLineTipHeight)
                     .text(date);
+
+                //add a background to it
+                var boxX = focus.select('[name=crossLineTipX]').node().getBBox();
+                focus.select('[name=crossLineTipRectX]')
```
xizi-xu: Maybe create a function for this instead of having the same code twice?

ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -518,7 +584,7 @@ angular.module('argus.directives.charts.lineChart', [])
                 setGraph(); //set up the chart
                 updateGraph(currSeries); //refill the data draw the line
                 addOverlay();
-
+                reset();
```
xizi-xu: Why do we need a reset for brush after resizing? It seems that the brush adjusts its size accordingly for resizing.

cnardi-dev: I believe its to remove the brush extent layer on initial load, to allow the user to create a new brush as a first use-case.  My of this file has the comments '//to remove the brush cover first for user the drag'.

prestonfff: Yes, it is to avoid the full brush. Just modified the code to achieve that.

ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -44,13 +44,33 @@ angular.module('argus.directives.charts.lineChart', [])
         }],
         // compile: function (iElement, iAttrs, transclude) {},
         link: function (scope, element, attributes) {
+            // scope.$watch('series', function() {
```
xizi-xu: Please delete line 47-66

PR #282, opened 2016-11-14
Finished graph toggling feature

xizi-xu:
- refactor the mouse over circle code 
- add graphClassName to each metric to bind all the graph elements together under one class


cnardi-dev:
Looks good!

PR #283, opened 2016-11-15
Truncate field names before sending a message to GOC.

bsura:


bsura:
@rajsarkapally-sfdc @dilipdevaraj-sfdc Please review.

PR #285, opened 2016-11-15
Improvements on yAxis rescaling and fixes on main chart brush problems

prestonfff:
- Add buffer above/below y Domain's max/min to make chart look less full
- Fix an issue that user cannot enable scroll when mainchart brush enables
- Fix an issue that crossline disappears when mainchart brush enables


cnardi-dev:
good fixes!

PR #286, opened 2016-11-16
Add Primary Key Auto Generation Annotation for History entity.

bsura:


bsura:
@dilipdevaraj-sfdc I have added primary key auto generation. This should fix your problem.

PR #288, opened 2016-11-16
Fixing resizing only affects last chart problem

prestonfff:


ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -881,7 +882,18 @@ angular.module('argus.directives.charts.lineChart', [])
             }
 
             // call resize when browser size changes
-            d3.select(window).on('resize', resize);
+            var parent = scope.$parent.$parent.$parent;
```
cnardi-dev: I think this is as good a setup as we could find.  So that we cover our bases, lets make sure this doesn't cause issues with other elements on the page, or any other quirky behavior.  Lets check FF and Safari browsers as well.

PR #292, opened 2016-11-18
Incr. wildcard expansion limit to 25, throw an exception when limit is exceeded, config changes for asynchbase

bsura:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/DefaultDiscoveryService.java
```diff
@@ -194,7 +192,8 @@ public DefaultDiscoveryService(SchemaService schemaService, SystemConfiguration
                                 }
                             } else {
                                 if (queries.size() == HARD_LIMIT) {
```
dilipdevaraj-sfdc: Should this check be if (queries.size() >= HARD_LIMIT)

ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/resources/MetricResources.java
```diff
@@ -157,10 +160,15 @@ public Response getAsyncMetricsJSON(@Context HttpServletRequest req,
...
 
         for (String expression : expressions) {
-            List<Metric> metricsForThisExpression = metricService.getMetrics(expression);
-
-            metrics.addAll(metricsForThisExpression);
+        	try {
+        		List<Metric> metricsForThisExpression = metricService.getMetrics(expression); 
+        		metrics.addAll(metricsForThisExpression);
+        	} catch(WildcardExpansionLimitExceededException e) {
+        		metricService.dispose();
```
dilipdevaraj-sfdc: If user has multiple expressions to be evaluated, and first expression results in WildcardExpansionLimitExceededException does it make sense to still proceed with the next set of expressions ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/DefaultDiscoveryService.java
bsura: If you take a closer look, == should suffice.

ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/resources/MetricResources.java
bsura: Yes. Totally. But we currently do not have a way to let the user know how many expressions succeeded and how many failed. For now, it probably makes sense to fail the entire operation.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/DefaultDiscoveryService.java
dilipdevaraj-sfdc: Agreed

PR #293, opened 2016-11-18
Add timeout for resize function and fix some empty data issue

prestonfff:


ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -878,7 +881,8 @@ angular.module('argus.directives.charts.lineChart', [])
 
             //extent, k is the least number of points in one line you want to see on the main chart view
             function setZoomExtent(k){
-              //TODO: deal with empty data
+                if(!currSeries || currSeries.length === 0) return;
```
xizi-xu: We probably should have a way to handle empty data initially instead of doing this check every time. @cnardi-dev

PR #296, opened 2016-11-19
Tooltip flipping and refactor some code

xizi-xu:
- Tooltip flips to the left side when there is not enough to display it on the right.
- Factor out graph element creation in `setGraphTools` function.
- Add some TODOs for empty data issues.

### DIFF
--- a/ArgusWeb/app/js/directives/charts/lineChart.js
+++ b/ArgusWeb/app/js/directives/charts/lineChart.js
@@ -296,7 +296,7 @@ angular.module('argus.directives.charts.lineChart', [])
                 tipItems = tip.append('g')
                     .attr('class', 'tooltip-items');
 
-                //focus tracking
+                //focus tracking/crossLine
                 crossLine = focus.append('g')
                     .attr('name', 'crossLine');
                 crossLine.append('line')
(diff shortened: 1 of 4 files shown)


ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -671,10 +679,13 @@ angular.module('argus.directives.charts.lineChart', [])
                     left: marginLeft};
 
                 //clear every chart
-                // BUG: resize causes charts to re-attach in wrong DOM position
+                //TODO: bug-resize causes charts to re-attach in wrong DOM position
                 d3.select(container).select('svg').remove();
```
cnardi-dev: I think this has been removed from previous 'resize' refactoring?

xizi-xu: I don't think so? @prestonfff Can you verify?

prestonfff: @xizi-xu @cnardi-dev It's still there and is used to remove all the svg first and create new ones based on the new size. The re-attach in wrong DOM position problem is fixed.

PR #303, opened 2016-11-28
Some Empty Dataset Handling

xizi-xu:
- Display a message when nothing is returned from TSDB or the expression does not exist.
- Show local timezone information in the time range label.

### DIFF
--- a/ArgusWeb/app/js/directives/charts/lineChart.js
+++ b/ArgusWeb/app/js/directives/charts/lineChart.js
@@ -95,15 +95,19 @@ angular.module('argus.directives.charts.lineChart', [])
             var height = parseInt((containerHeight - marginTop - marginBottom) * mainChartRatio);
             var height2 = parseInt((containerHeight - marginTop - marginBottom) * brushChartRatio) - brushHeightFactor;
 
-            var margin = {top: marginTop,
+            var margin = {
+                top: marginTop,
                 right: marginRight,
                 bottom: containerHeight - marginTop - height,
-                left: marginLeft};
+                left: marginLeft
+            };
 
-            var margin2 = {top: containerHeight - height2 - marginBottom,
+            var margin2 = {
+                top: containerHeight - height2 - marginBottom,
                 right: marginRight,
                 bottom: marginBottom,
-                left: marginLeft};
+                left: marginLeft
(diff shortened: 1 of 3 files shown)


ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -914,41 +941,39 @@ angular.module('argus.directives.charts.lineChart', [])
             }
 
             //extent, k is the least number of points in one line you want to see on the main chart view
-            function setZoomExtent(k){
-              //TODO: empty data issue
-                if(!currSeries || currSeries.length === 0) return;
-                var numOfPoints= currSeries[0].data.length;
+            function setZoomExtent(k) {
+                var numOfPoints = currSeries[0].data.length;
```
prestonfff: what if currSeries is []?

ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -974,21 +999,29 @@ angular.module('argus.directives.charts.lineChart', [])
...
-
-            // initialize starting point for graph settings & info
-            addOverlay();
-            updateDateRange();
-            enableBrushTime();
-            reset();    //to remove the brush cover first for user the drag
+            // create graph only when there is data
+            if (!series || series.length === 0) {
+                console.log("Empty data from chart data processing");
+                return;
```
prestonfff: So There will not be a empty chart place holder? Maybe a placeholder chart with "Empty data from chart data processing" info in the center?

prestonfff:
Maybe we can do a separate commit for all the reformatting of codes, just to make it faster to locate the real changes and make reviewing easier.

ArgusWeb/app/js/directives/charts/lineChart.js
xizi-xu: This function is only used in updateGraph. currSeries and series are checked for empty before updateGraph is called. No need to check for it again

ArgusWeb/app/js/directives/charts/lineChart.js
xizi-xu: This technically should never happen. It's just a place holder for now.

PR #304, opened 2016-11-28
Fix divs

ddjohnson:
@saaja-sfdc


saaja-sfdc:
@ddjohnson can you please resubmit this PR so that it is merging into /develop rather than /master.  You should be working of the /develop branch.

PR #312, opened 2016-12-02
Add local storage for user menu option and fix some small bugs

prestonfff:
- Add local storage for user menu option
- Fix some small issues on toggle and switching tabs


ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -1,10 +1,12 @@
 'use strict';
 
 angular.module('argus.directives.charts.lineChart', [])
-.directive('lineChart', ['$timeout', function($timeout) {
+.directive('lineChart', ['$timeout', 'Storage', '$routeParams', function($timeout, Storage, $routeParams) {
```
xizi-xu: There was an inputTracker service you could use if you want.

ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -61,23 +63,38 @@ angular.module('argus.directives.charts.lineChart', [])
...
+                isTooltipOn : true
+            };
+
+            scope.dashboardId = $routeParams.dashboardId;
+            
+            var menuOption = Storage.get('menuOption_' + scope.dashboardId +'_' + lineChartIdName + scope.lineChartId);
+            if(menuOption){
+                    scope.menuOption = menuOption;
+            }
+
```
xizi-xu: I recommend line 80-93 to be moved to the controller of the directive. Here are the reasons:
1. The controller is called before the link and this ensures these variables got initiated before the DOM is populated. 
2. Variables in the controller can be used by other used by other directives, so things will be more flexible.
3. Scope manipulation should be done by the controller.

cnardi-dev: Good points here.  We can move forward with these changes in the next PR.

ArgusWeb/app/js/directives/charts/lineChart.js
prestonfff: @xizi-xu  I see that. Just for this simple case, I don't need the default value for the get and set, so I just use the Storage and save some function call.

PR #327, opened 2016-12-05
Breadcrumbs & empty data check enhancement 

xizi-xu:
- Add 'use strict' to prevent some strange behavior of breadcrumbs redirect.
- Remove metrics that cannot be graph from series prior to generating the chart.


xizi-xu:
@cnardi-dev When I use the breadcrumbs to go back to dashboard list view, it triggers the API query call, which should not happen because of the session storage. I added 'use restrict' and things seem to work just fine now. It probably is just some strange javascript thing.

cnardi-dev:
OK, understood.  I haven't known of any issues with not using 'use strict', but it is good practice anyway.  Lets make sure to test this again once its on a real QA server.  Will need to determine if it's an issue on Alerts as well.

PR #330, opened 2016-12-07
D3 loading spinner for charts

prestonfff:


ArgusWeb/app/js/templates/charts/topToolbar.html
```diff
@@ -66,4 +66,6 @@ <h3 class="chartTitle">{{chartConfig.chartTitle}}</h3>
 
 		<div class="toolbarItem right dateRange" id="date-range-{{chartConfig.chartId}}">{{dateRange}}</div>
 	</div>
+	<div ng-loading="graphRendered"></div>
```
xizi-xu: This may not be needed because d3 rendering is very fast.

prestonfff:
@cnardi-dev It should work independently.

ArgusWeb/app/js/templates/charts/topToolbar.html
prestonfff: I notice a lag for 1s after the loading spinner stops spinning. Might not be caused by the rendering but I am still investigating the cause.

prestonfff: Fixing it. Might modify the place to put graphRendered, since I guess d3 rendering are all done by callback and when graphRendered is set to true, d3 might be still rendering.

PR #332, opened 2016-12-07
fix bug: seriesname and seriescolor

cnardi-dev:
$promise + closure = doh!

When more time is available, should refactor further: 

- move api calls to service which will return data inside $promise with closure.  This will avoid having to call another method with many params to create closure.

Refer to "js/services/teams.js" in the Driven project for solid example.


xizi-xu:
For some reason, this is really slow for me. Not sure if it's a code issue, network issue, or ws host issue for the moment.

PR #333, opened 2016-12-07
Move sonatype deployments into a profile.

rmelick:
We need to set up publishes of argus into our local nexus in case we have urgent bugfixes.

The sonatype-staging profile allows sonatype deployments to be deactivated during our build job, which needs to publish to a different nexus.  It is active by default, so this should not change any of the existing build scripts.
Also, I set specific version of maven-deploy plugin to allow overriding repositories (the correct parameters to override the release and snapshot repositories with different urls were only available in 2.8).


bsura:
@saaja-sfdc Can you please review this?

PR #345, opened 2016-12-09
Render only data within the time domain in path element and small fix

prestonfff:
- Fix rendering error when a path is outside scope
- Render only data within the time domain in path element (rely on clip-path before)


xizi-xu:
I do not understand the error about rendering error when a path is outside of scope. Please leave this unmerged until we discuss this.

prestonfff:
@xizi-xu I explained it on the card on kanban

PR #347, opened 2016-12-09
Notification cooldown per trigger,metric combination

bsura:
1. Notification cooldown now associated with a Trigger,Metric combination.
2. Refactor alert service test. Replace AlertServiceIT with unit tests using Mockito.
3. Update pom.xml dependencies.


bsura:
Please do not merge this branch until it has been properly reviewed.

Issue #342, opened 2016-12-08
Runtime configuration of database connection 


rmelick:
I would like to be able to configure the database connection (at a minimum the user name and password) at runtime.  This would allow me to deploy my same built war to multiple environments (for example test and production).

It seems there are a few comments out there about how to do it for hibernate or through directy property setting in java (http://stackoverflow.com/questions/8324821/properties-reference-for-hibernate-in-persistence-xml and http://stackoverflow.com/questions/3935394/how-to-externalize-properties-from-jpas-persistence-xml).  How do you do it at salesforce?

saaja-sfdc:
it's set in the argus-build.properties when we build the artifacts for production.  These properties act as a resource filter that ultimately updates persistence.xml in argus-core.  The property name is build.property.persistence.unit and should be updated with the content of the <provider/> element of your specific persistence unit engine.

rmelick:
@saaja-sfdc , do you mind explaining how the build and deployment process for Argus works at Salesforce?  Requiring a separate build for each environment won't work well with creating Docker images, and it also won't match with how we were hoping to build and deploy at kiwigrid.  

Our plan was as follows:

- In the normal case, we would be deploying the publicly available argus artifacts from maven central or from the public docker images.  If we had a bugfix or feature that was not time critical, we would create a pull request against the develop branch in the public repo and wait for the new release.
- For super urgent hotfixes, we would fork the public repo and create a new branch based on our currently deployed release (e.x. 2.2).  We would make the patch there, and also create a pull request to the public develop branch.  While we waited for the new public release, we would run a local build of our fork, push the artifacts to our internal nexus, and deploy them.  Then, once the new public version was available, we would deploy it instead of our hotfixed version.

Obviously, this won't work if all the public builds are environment specific.  It seems like the persistence.xml is the maing thing preventing us from setting all the configuration at runtime instead of build time.  I think we could fix that by modifying the configurePersistence() method in SystemInitializer to allow setting properties on the jpaPersistModule.

Does it make sense to try and use the same argus build in multiple environments, or am I completely missing something?

saaja-sfdc:
No you're not missing anything.  This is our goal as well.  The static binding of JPA at build time is an artifact from early Argus.  

👍

PR #348, opened 2016-12-09
Allow overriding of jpa properties by using system.property.jpa prefix

rmelick:
Issue #342 : This should allow us to use the same built war in multiple environments by specifying the database connection properties at runtime.

### DIFF
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/system/SystemInitializer.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/system/SystemInitializer.java
@@ -74,6 +74,7 @@ import java.util.Properties;
  * @author  Tom Valine (tvaline@salesforce.com), Bhinav Sura (bhinav.sura@salesforce.com)
  */
 final class SystemInitializer extends AbstractModule {
+    private static final String JPA_PROPERTY_PREFIX = "system.property.jpa";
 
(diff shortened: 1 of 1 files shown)


ArgusCore/src/main/java/com/salesforce/dva/argus/system/SystemInitializer.java
```diff
@@ -175,7 +176,22 @@ private void configureSystem() {
...
+        JpaPersistModule jpaPersistModule = new JpaPersistModule("argus-pu");
+        jpaPersistModule.properties(getJpaProperties(_systemConfiguration));
+        binder().install(jpaPersistModule);
+    }
+
+    private Properties getJpaProperties(Properties properties) {
+        Properties jpaProperties = new Properties();
+        for (Object rawPropertyNameWithPrefix : properties.keySet()) {
+            String propertyNameWithPrefix = (String) rawPropertyNameWithPrefix;
+            if (propertyNameWithPrefix.startsWith(JPA_PROPERTY_PREFIX)) {
```
rmelick: I would love some comments about whether having a property prefix like this fits with the rest of the configuration design.  It didn't seem like it made sense to define a service and fixed set of properties for this, since they are always just forwarded to the jpa module.

saaja-sfdc:
👍

PR #349, opened 2016-12-10
Misplacement of series appending.

xizi-xu:
Empty returned metrics error message was not displaying because it was not added to the series to pass into graph rendering.


cnardi-dev:
Is this CI check due to failed unit tests?  If not, we can merge this.

xizi-xu:
@cnardi-dev I'm not sure. Waiting on Pimeng to take a look.

prestonfff:
The error report seems have nothing to do with the UI part.

PR #359, opened 2016-12-16
Add extra log for metrics pushed from monitoring service. If hashmap …

dilipdevaraj-sfdc:
…size copy, when trying to push metrics are unequal , then log warnign message.

### DIFF
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DefaultMonitorService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DefaultMonitorService.java
@@ -616,7 +616,7 @@ public class DefaultMonitorService extends DefaultJPAService implements MonitorS
     private class MonitorThread extends Thread {
 
         /**
-         * Creates a new SchedulingThread object.
+         * Creates a new MonitorThread object.
          *
          * @param  name  The thread name.
          */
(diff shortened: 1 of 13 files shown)


ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DefaultMonitorService.java
```diff
@@ -639,15 +639,25 @@ public void run() {
...
 
         private void _pushCounters() {
+        	int sizeJVMMetrics = 0;
             _logger.debug("Pushing monitor service counters for {}.", HOSTNAME);
 
             Map<Metric, Double> counters = new HashMap<>();
-
+            
             _updateJVMStatsCounters();
+            sizeJVMMetrics = _metrics.size();
```
bsura: This should go inside of synchronized block. Since metrics may be added to _metrics after you've instantiated sizeJVMMetrics.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DefaultMonitorService.java
```diff
@@ -658,6 +668,7 @@ private void _pushCounters() {
                 entry.getKey().setDatapoints(dataPoints);
             }
             if (!isDisposed()) {
+            	_logger.info("Pushing {} monitoring metrics to TSDB.", counters.size());
```
bsura: Since this is only called every minute, I would log the entire counters map so that we can debug at a later point in time.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DefaultMonitorService.java
dilipdevaraj-sfdc: Done

PR #376, opened 2017-01-05
Do not show graph menu options if no graph renders

prestonfff:


cnardi-dev:
Need to update this PR with latest changes from develop.  Once this is merge, I will re-deploy to QA instance. Thnx.

PR #383, opened 2017-01-10
Create shared alert UI component

xizi-xu:
1. Alert list page now has personal and shared alerts tabs
2. An Alert can be marked as shared in the alert detail page
3. Add UI authorization check for shared alerts (view only)
3. Create a new service for item list manipulation
4. Create a new directive for tabs list


xizi-xu:
There is a bug that alerts appear twice when they are shared and you are the owner.

cnardi-dev:
Closing this to update and reopen on share-alert branch.

PR #385, opened 2017-01-11
Add capability to Alias scope, users can now subtract time from agDate controls.

bsura:


bsura:
For e.g. ALIAS(expr, #new_metric#, #literal#, #new_scope#, #literal#)

PR #393, opened 2017-01-16
rescale y axis when hiding sources from legend

prestonfff:


ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -460,8 +462,8 @@ angular.module('argus.directives.charts.lineChart', [])
                 var mouseY = y.invert(positionY);
 
                 if(isBrushInNonEmptyRange()) {
-                    currSeries.forEach(function (metric) {
-                        if (metric.data.length === 0) {
+                    currSeries.forEach(function (metric, index) {
+                        if (metric.data.length === 0 || !scope.sources[index].displaying) {
```
cnardi-dev: Should we check for !scope.sources as well ?  What if the chart has no sources, is this possible ?

cnardi-dev: Actually this should be ok since we're inside the look for series.

PR #401, opened 2017-01-24
Fix the bug of loading spinner stuck on viewmetrics page

xizi-xu:


xizi-xu:
@bsura Yea, I can do that.

PR #411, opened 2017-01-28
Change datapoints map from Map<Long, String> to Map<Long, Double>.

bsura:


ArgusCore/src/test/java/com/salesforce/dva/argus/service/DefaultAlertServiceTest.java
```diff
@@ -108,7 +108,7 @@ public void testExecuteScheduledAlerts_ForOneTimeSeriesMultipleTriggers() {
 		dps.put(1000L, "11");
 		dps.put(2000L, "21");
 		dps.put(3000L, "31");
-		metric.setDatapoints(dps);
+		metric.setDatapoints(_convertDatapoints(dps));
```
dilipdevaraj-sfdc: For completeness sake, can we change even in tests the dps map to be  Map<Long, Double> 
so that we need not have additional method _convertDatapoints

bsura:
@dilipdevaraj-sfdc Nope. We no longer use those Transforms.

ArgusCore/src/test/java/com/salesforce/dva/argus/service/DefaultAlertServiceTest.java
bsura: I did not want to manually go through the process of changing Strings like "1.0" to Doubles 1.0. Hence I wrote a utility method and used it everywhere.

PR #415, opened 2017-01-31
Y-Axis buffer for straight line graphs 

xizi-xu:
- Add 1 scale at the bottom and 2 scales at the top for graphs that have single range 
- Add white background for annotation label
screen shot 2017-01-31 at 10 24 35 am


cnardi-dev:
Looks good. Will hold off on merging into develop until we can get QA server back up today or tomorrow.

PR #421, opened 2017-02-03
Store metrics using Phoenix

twdsilva:
@bsura


bsura:
There are still compilation errors. Please rebase against the salesforce/Argus/develop and then try to resubmit. We recently changed datapoints Map from <Long, String> to <Long, Double>.

twdsilva:
@bsura  The build passes with this patch now.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/tsdb/PhoenixTSDBEngine.java
```diff
@@ -0,0 +1,248 @@
...
+	
+	private String getPhoenixQuery(MetricQuery query) {
+		
+		String viewName = getPhoenixViewName(query.getScope(), query.getMetric());
+		String agg = convertArgusAggregatorToPhoenixAggregator(query.getAggregator());
+		
+		String tagkeys = "", tagWhereClaue = "";
+		for(Map.Entry<String, String> tagEntry : query.getTags().entrySet()) {
+			tagkeys += ", \"" + tagEntry.getKey() + "\"";
+			//TODO: Add support for tagKey=* and tagKey=a|b
```
bsura: Can you finish this TODO so that we have parity with what OpenTSDB provides. It should be straight forward.

PR #423, opened 2017-02-03
Refactor Method Names. 

bsura:


bsura:
@dilipdevaraj-sfdc Refactored method names.

PR #437, opened 2017-02-10
Add ALIASBYTAG Transform.

bsura:
E.g. ALIASBYTAG(-30d:argus.jvm:mem.heap.used{host=*}:zimsum:1d-sum, #host#) will set the displayName for the metric with the value of the host.


bsura:
@cnardi-dev @xizi-xu Please verify if dataProcessing.js is the only file which set the name of the series. If it's set in some other place as well, then we will need to make similar changes there too.

cnardi-dev:
Thanks @bsura, we will check out this feature and review the UI setup during next QA cycle.

PR #440, opened 2017-02-11
Fix the bug of status indictor using string comparison instead of number

xizi-xu:


ArgusWeb/app/js/services/charts/chartRendering.js
```diff
@@ -48,15 +48,19 @@ angular.module('argus.services.charts.rendering', [])
...
-								$('#' + attributes.name + '-status').removeClass('red orange green').addClass('red');
-						} else if (lastStatusVal > attributes.lo && lastStatusVal < attributes.hi) {
-								$('#' + attributes.name + '-status').removeClass('red orange green').addClass('orange');
-						} else if (lastStatusVal > attributes.hi) {
-								$('#' + attributes.name + '-status').removeClass('red orange green').addClass('green');
-						}
+                    var lastStatusValNum = Number(lastStatusVal);
+                    var lowValNum = Number(attributes.lo);
+                    var highValNum = Number(attributes.hi);
+
```
bsura: Although this wouldn't generate any errors, I would be nice to log to console if attributes.lo or attributes.hi is undefined. That way one can easily check the console and figure out that hi and lo are undefined. Makes sense?

xizi-xu: Do we want to have a growl message as well when lo or high is not defined?

xizi-xu: @bsura

bsura: Umm. Growl would be too much I suppose. If it were me, I would probably just console.log() it for the purpose of debugging. I would leave that up to you.

PR #455, opened 2017-02-22
Change hash from md5() to String.hashcode()

bsura:


bsura:
@rajsarkapally-sfdc @dilipdevaraj-sfdc

PR #462, opened 2017-02-25
Add Severity Level to GOC Notification

dilipdevaraj-sfdc:
@rajsarkapally-sfdc @bsura 

Can you guys please review ?


ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Notification.java
```diff
@@ -86,6 +86,8 @@
     )
     List<Trigger> triggers = new ArrayList<>(0);
     boolean isSRActionable = false;
+    int severityLevel = 5;
```
dilipdevaraj-sfdc: Decided not to to use the denormalized GOC_Notification_Info table since some of the fields for GOC can be reused for other notification types as well. This would be zero or one to one mapping, so notification reads need not have a join but can be obtained from notification table itself. There would be some whole scale changes to Notification if we decide to pursue this approach. Hence limiting these changes for next release to adding GOC severity field only instead of refactoring.

ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/dto/NotificationDto.java
```diff
@@ -271,7 +272,29 @@ public boolean getSRActionable() {
...
+    /**
+     * Sets the severity level of notification
+     *
+     * @param  severityLevel  The severity level
+     * @throws  IllegalArgumentException  If an error occurs.
+     */
+    public void setSeverityLevel(int severityLevel) {
+        if (severityLevel < 1 || severityLevel > 5) {
+            throw new IllegalArgumentException("The severty level should be between 1-5");
+        }
```
bsura: This should ideally be checked in ArgusCore/Notification.java

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Notification.java
```diff
@@ -371,6 +372,24 @@ public void setSRActionable(boolean isSRActionable) {
...
+        return severityLevel;
+    }
+
+    /**
+     * Sets the severity level of notification
+     *
+     * @param  severityLevel  The severity level
+     */
+    public void setSeverityLevel(int severityLevel) {
+        this.severityLevel = severityLevel;
```
bsura: Add the check here.

ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/dto/NotificationDto.java
```diff
@@ -61,6 +61,7 @@
     private List<BigInteger> triggersIds = new ArrayList<>();
     private BigInteger alertId;
     private String customText;
+    private int severityLevel;
```
bsura: Not sure if would affect anything but if Notification entity has a default setting of 5, then the DTO should also have a default setting of 5.

ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/dto/NotificationDto.java
dilipdevaraj-sfdc: Done

ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/dto/NotificationDto.java
dilipdevaraj-sfdc: Done

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Notification.java
dilipdevaraj-sfdc: Done. Pls review.

PR #464, opened 2017-02-27
Kiwigrid authentication

axdotl:


bsura:
Let's put this on hold for a while until we merge Token Based Auth. That might change this impl a bit.

PR #470, opened 2017-02-28
Fix an issue that annotation hiding does not work

prestonfff:


cnardi-dev:
This PR seems to be stuck. Maybe a new push will get the checks to complete?

PR #472, opened 2017-02-28
DEDUCT TRANSFORM

aertoria:
DEDUCT TRANSFORM
Refer to Issue#375
@bsura @saaja-sfdc 

DEDUCT(M1, M2) It returns M1 - M2 from time dimension. It remove all items from M1 whose keys are found in M2. This is the opposite of UNION.

Note: it is not MIN, MIN only align both Metric up by key, and do the value minus.

Usage: DEDUCT(List).
Example:
input:
M1 is 1L:10, 2L:10, 3L:20
M2 is 1L:10

output as DEDUCT(M1,M2)
M is 2L:10, 3L:20


.gitignore
```diff
@@ -12,6 +12,7 @@ argus-build.properties
 .settings/
 .DS_Store
 /config
+*.properties
```
bsura: Can you not remove .gitignore from the commit?

bsura:
@aertoria PR failed with Compilation Error. Can you fix those and resubmit? Also remove .gitignore from the commit.

PR #483, opened 2017-03-07
Improvements and bug fixes for SchemaService.

bsura:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AsyncHbaseSchemaService.java
```diff
@@ -637,22 +675,52 @@ public Void call(Exception arg) throws Exception {
...
-    	_put(tableName, rowKeyStr);
+    private void _putWithoutTag(Metric metric) {
+    	String rowKeyScopeTable = _constructRowKey(metric.getNamespace(), metric.getScope(), metric.getMetric(), 
+    			null, null, SCOPE_SCHEMA_TABLENAME);
+    	String rowKeyMetricTable = _constructRowKey(metric.getNamespace(), metric.getScope(), metric.getMetric(), 
+    			null, null, METRIC_SCHEMA_TABLENAME);
+
+    	boolean exists = false;
+    	if(_cacheEnabled) {
+    		exists = _trie.getValueForExactKey(rowKeyScopeTable) != null;
```
dilipdevaraj-sfdc: Right now we are using trie only for writing(by using it as read before write).
Do you think it will be worthwhile to also use if for read path where we can use it for prefix matching, instead of going to Hbase ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AsyncHbaseSchemaService.java
```diff
@@ -367,18 +380,18 @@ public Object call(ArrayList<ArrayList<KeyValue>> rows) throws Exception {
      * 
      * For e.g. suppose if we have the following rows in HBase:
      * 
-     * scope:metric1:null:null:null
-     * scope:metric2:null:null:null
+     * scope\0metric1\0$\0$\0$
```
bsura: Yes. But that can break the logic. The whole point of using \u0000 as the separator is that it's ASCII value is 0 and so does not mess with the lexicographical order.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AsyncHbaseSchemaService.java
bsura: A couple of microseconds.

bsura: Umm. Holding a trie on the web server would consume a lot of memory. Hence I chose not to solve for that usecase.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AsyncHbaseSchemaService.java
```diff
@@ -637,22 +675,52 @@ public Void call(Exception arg) throws Exception {
...
+
+    	boolean exists = false;
+    	if(_cacheEnabled) {
+    		exists = _trie.getValueForExactKey(rowKeyScopeTable) != null;
+    	}
+    	
+    	if(!exists) {
+    		_put(SCOPE_SCHEMA_TABLENAME, rowKeyScopeTable);
+    		_put(METRIC_SCHEMA_TABLENAME, rowKeyMetricTable);
+    		if(_cacheEnabled) {
```
bsura: I don't have an exact no. of write latency but it should be of the same order as reads.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AsyncHbaseSchemaService.java
```diff
@@ -80,38 +88,43 @@
...
+	
 	private static String METRIC_SCHEMA_TABLENAME;
 	private static String SCOPE_SCHEMA_TABLENAME ;
     private static final byte[] COLUMN_FAMILY = "f".getBytes(Charset.forName("UTF-8"));
     private static final byte[] COLUMN_QUALIFIER = "c".getBytes(Charset.forName("UTF-8"));
-    private static final byte[] CELL_VALUE = "1".getBytes(Charset.forName("UTF-8"));
-    private static final char ROWKEY_SEPARATOR = ':';
-    private static final char[] WILDCARD_CHARSET = new char[] { '*', '?', '[', ']', '|' };
+    private static final byte[] CELL_VALUE = new byte[1];
+    private static final char ROWKEY_SEPARATOR = '\u0000';
```
dilipdevaraj-sfdc: maybe you can add the explanation of ROWKEY_SEPARATOR being \u0000
so the lexicographical ordering is maintained
Row 1 scope1:metric
Row 2 scope:metric

While new rowkey will have right lexicographical ordering
Row 1 scopemetric
Row 2 scope1metric

PR #501, opened 2017-03-17
Cache users

bsura:


bsura:
@dilipdevaraj-sfdc @rajsarkapally-sfdc Implementing a user cache that holds the 50 most frequently used user objects in memory. Please review.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/jpa/DefaultUserService.java
```diff
@@ -112,6 +144,7 @@ public void deleteUser(PrincipalUser user) {
 
         deleteEntity(em, user);
         em.flush();
+        USERCACHE.invalidate(user.getUserName());
```
rajsarkapally-sfdc: @bsura Let user1 is deleting user_x. Let the code executed until line 146. Now if user2 is trying to read user_x, the user_x object exist in cache so it will be returning from cache. so how about invalidating cache for user_x before user_x is deleted from DB?

bsura: Makes sense.

PR #519, opened 2017-03-30
Change null string placeholder from dollar to hash

bsura:


bsura:
@dilipdevaraj-sfdc @rajsarkapally Please approve.

PR #534, opened 2017-04-26
Add ag-select directive

prestonfff:
- similar use with ag-text and other control directive
- support keyword search/auto completion
- generate a scroll bar when options are of large number
- use option tag inside it to define the options text and values

Will add support for infinite scroll when using api to add the options for scenarios like filling the field in metric discovery.

### DIFF
--- a/ArgusWeb/app/js/argus.js
+++ b/ArgusWeb/app/js/argus.js
@@ -46,9 +46,9 @@ angular.module('argus', [
   'argus.directives.dashboardResource',
   'argus.directives.controls.dashboard',
   'argus.directives.controls.date',
-  'argus.directives.controls.dropdown',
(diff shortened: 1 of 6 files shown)


ArgusWeb/app/js/argus.js
```diff
@@ -202,9 +202,10 @@ require("./directives/breadcrumbs");
 require("./directives/dashboardResource");
 require("./directives/controls/dashboard");
 require("./directives/controls/date");
-require("./directives/controls/dropdown");
+require("./directives/controls/select");
 require("./directives/controls/submit");
 require("./directives/controls/text");
+require("./directives/controls/select");
```
xizi-xu: You have this twice?

PR #554, opened 2017-05-05
Skip datapoints that contain null value instead of erroring out

bsura:


bsura:
@kgowdru Please review. We will skip serializing datapoints that contain null values.

PR #560, opened 2017-05-12
Set severity level and sractionable on create notification in web ser…

dilip-devaraj:
…vice resources


dilipdevaraj-sfdc:
@rajsarkapally-sfdc  Can you please review

PR #568, opened 2017-05-23
Add multiple yAxis feature

prestonfff:
- Add multiple yAxis feature, with attribute "extra-yAxis" in <ag-metric> directive to define extra yAixs


ArgusWeb/app/js/config.js
```diff
@@ -22,7 +22,7 @@
 
 angular.module('argus.urlConfig', [])
 .constant('CONFIG', {
-	version: '2.7.1',
+	version: '2.6.0',
```
xizi-xu: Please change this to 2.71 as that's the newest version

PR #593, opened 2017-07-06
Update AuthService login() method. Should never throw TokenExpiredException.

bsura:


ArgusSDK/src/main/java/com/salesforce/dva/argus/sdk/AuthService.java
```diff
@@ -76,7 +76,7 @@
      * @throws TokenExpiredException
```
dilipdevaraj-sfdc: Remove from java doc too.

bsura: Done.

PR #595, opened 2017-07-07
Only retrieve metadata columns from DB for APIs that returns a collection of entities

bsura:
- Retrieve metadata columns from DB for APIs that returns a collection of entities
- Allow users to use UNION constant for a few more transforms where it makes sense
- Minor code refactoring


bsura:
@dilipdevaraj-sfdc @rajsarkapally-sfdc Please review when you get a chance.

PR #597, opened 2017-07-11
fix an issue of stack area chart not syncing

prestonfff:


xizi-xu:
This is a duplicate of #598

PR #599, opened 2017-07-13
Metric Scanner Implementation

adechambers:


dilipdevaraj-sfdc:
@adechambers
Can you resolve the conflicting file so unit tests can run on travis.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/AbsoluteValueMapping.java
```diff
@@ -56,11 +58,30 @@
...
         return absDatapoints;
     }
+	
+    @Override
+    public Map<Long, Double> mappingScanner(MetricScanner scanner) {
+    		Map<Long, Double> absDatapoints = new HashMap<>();
+    		
+    		synchronized (scanner) {
+	    		while (scanner.hasNextDP()) {
+	    			Map.Entry<Long, Double> nextDP = scanner.getNextDP();
```
saaja-sfdc: I like the use of Map.Entry here.  Nicely done.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/AbstractArithmeticTransform.java
```diff
@@ -86,6 +88,88 @@
...
+    	
+    	Metric result = new Metric(getResultScopeName(), RESULT_METRIC_NAME);
+    	Map<Long, Double> resultDatapoints = new HashMap<>();
+    	List<MetricScanner> remainingScanners = scanners.subList(1, scanners.size());
+    	List<Map<Long, Double>> datapoints = new ArrayList<>(remainingScanners.size());
+    	for (int i = 0; i < remainingScanners.size(); i++) {
+    		datapoints.add(new HashMap<>());
+    	}
+    	
+    	synchronized(scanners.get(0)) {
```
saaja-sfdc: Does it make sense to just use a synchronized block to create copies of the maps you need to access and then you can traverse the copies without the need for future synchronization? It would reduce the time you need exclusive access to the scanner list, since you'd be working with a copy of it.

new List<MetricScanner> scannerCopies;
synchronized(originalMap) {
  scannerCopies = new ArrayList(scanners);
}

while (...) {
  // do something on the copies of scanners
}

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/AnomalyDetectionGaussianTransform.java
```diff
@@ -77,12 +78,52 @@
         resultMetrics.add(predictionsNormalized);
         return resultMetrics;
     }
+    
+    @Override
+    public List<Metric> transformScanner(List<MetricScanner> scanners) {
+    	if (scanners == null) {
```
saaja-sfdc: For these, is there a method System.assert(condition, message, exceptionType) that you can use rather than constructing a new exception within an if/else?

saaja-sfdc: See the following code on this link for an example:  https://github.com/salesforce/Argus/pull/599/files?diff=unified#diff-a38764d130ea8dc982ead8c2e8c7a07bR131

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/FillTransform.java
```diff
@@ -236,7 +332,12 @@ public String getResultScopeName() {
 
     @Override
     public List<Metric> transform(List<Metric>... listOfList) {
-        throw new UnsupportedOperationException("Fill doesb't need list of list!");
+        throw new UnsupportedOperationException("Fill doesn't need list of list!");
```
saaja-sfdc: nice catch on the spelling error.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/AbstractArithmeticTransform.java
adechambers: Are you suggesting that I would instead used synchronized blocks on the copies of these original scanners? I was thinking that I needed to synchronize the blocks where I iterated through the scanners in order to make sure that hasNextDP() and getNextDP() were executed without anything else happening to that scanner between those lines. So would I then just synchronize the blocks iterating through the copied scanners but keep the original scanner available for more operations?

ArgusCore/src/test/java/com/salesforce/dva/argus/service/MetricScannerTest.java
```diff
@@ -0,0 +1,1042 @@
...
+	}
+	
+	private void checkDatapointMatch(Map<Long, Double> expected, MetricScanner ms) {
+		Map<Long, Double> actual = new TreeMap<Long, Double>();
+		while (ms.hasNextDP()) {
+			Map.Entry<Long, Double> dp = ms.getNextDP();
+			actual.put(dp.getKey(), dp.getValue());
+		}
+		
+		//System.out.println(actual + " is actual");
```
dilipdevaraj-sfdc: Please remove unused commented code.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/tsdb/MetricScanner.java
```diff
@@ -0,0 +1,516 @@
...
+	 */
+	private void asyncFetch() {
+		if (pointer != null && lastEndTimestamp.get() < query.getEndTimestamp()) {	
+			if (pointer.get() > Math.round((FETCH_PERCENTAGE * (lastEndTimestamp.get() - lastStartTimestamp.get()))) + lastStartTimestamp.get()) {
+				fetch();
+			}
+		}
+		else {
+			if (!query.getEndTimestamp().equals(lastEndTimestamp.get())) {	
+				fetch();
```
dilipdevaraj-sfdc: This method as discussed is used in a synchronous() fashion. Not sure if you intend to do this in a separate thread

dilipdevaraj-sfdc:
Right now only the unit tests are using the new MetricScanner implementation.
The transforms invoked through the WebService are still using the existing getMetrics() call. We can discuss with team if we plan on adding a second option of also invoking the MetricScanner implementation through web service
(See MetricReader.jj)

ArgusCore/src/main/java/com/salesforce/dva/argus/service/tsdb/DefaultTSDBService.java
```diff
@@ -351,6 +351,53 @@ public void putMetrics(List<Metric> metrics) {
...
+    
+    		Map<MetricQuery, List<Metric>> metricMap;
+    		MetricQuery miniQuery;
+    		Long startTime;
+    		Long stopTime;
+    		do {
+    			
+    			additionalChunks++;
+	    		startTime = query.getStartTimestamp() + chunkTime * additionalChunks;
+	    		stopTime = Math.max(query.getEndTimestamp(), startTime + chunkTime);
```
dilipdevaraj-sfdc: What is the logical reasoning for Math.max here?
endTimestamp will be always greater than start+chunk time, except for maybe last iteration.
If this is chunk of 1 hr in a 24 hr period are you intending start - 0, stop -1, start -1, stop-2 etc ?

adechambers: That check was in place for the last iteration. I didn't want to have any roundoff error that could potentially return a datapoint for just outside of the query range. In terms of how the chunks are split up with start and stop times, that was my intention.

dilipdevaraj-sfdc: But the effect of this would be for all iterations except the last one, since Math.max is used stopTime will always be query.getEndTimestamp(). Is that the intention ?

adechambers: I intended for this to set stopTime to startTime + chunkSize for all iterations except for the last one where it is set to query.getEndTimestamp() in case startTime + chunkSize would seep past this boundary.

dilipdevaraj-sfdc: In that case shouldnt it be Math.min ?

adechambers: Yep! Thanks.

dilipdevaraj-sfdc:
Right now the MetricScanner is invoked only with Argus transfroms. But what if the user performs plain TSDB queries - shouldn't the MetricScanner be invoked for this to ?
The bottleneck is not the transforms but the time to get data from underlying datastore.

PR #605, opened 2017-07-19
Fix download not working in bundle file

prestonfff:


xizi-xu:
The webpack bundled version worked with Chrome and Safari. Anyone tested with FF?

PR #630, opened 2017-08-15
Add load spinner for dashboard&alert history

prestonfff:


cnardi-dev:
Agreed. Can add in next PR for Alerts evaluation log tab.

PR #632, opened 2017-08-18
Fix ag-table shows metric name instead of alias name

prestonfff:


ArgusWeb/app/js/services/agTableService.js
```diff
@@ -50,7 +26,7 @@ angular.module('argus.services.agTableService', [])
 				for (i in data) {
 					dps = data[i].datapoints;
```
xizi-xu: `dps` is not defined in this loop

PR #635, opened 2017-08-28
Catch all exception in a few redis cache methods so the caller code c…

dilipdevaraj-sfdc:
…an bypass cache and get from underlying store

### DIFF
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/cache/RedisCacheService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/cache/RedisCacheService.java
@@ -28,7 +28,7 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
-	 
+
 package com.salesforce.dva.argus.service.cache;
 
 import com.google.inject.Inject;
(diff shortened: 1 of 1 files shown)


ArgusCore/src/main/java/com/salesforce/dva/argus/service/cache/RedisCacheService.java
```diff
@@ -63,285 +63,293 @@
...
+	//~ Methods **************************************************************************************************************************************
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public <V> V get(String key) {
+		V returnValue = null;
+		try{
+			returnValue = (V) _jedisClusterClient.get(key);
+		}catch (Exception ex) {
+			_logger.error("Exception in cache service: {} ", ex.getMessage());
```
dilipdevaraj-sfdc: Catch exception here. Rest of all is due to indentation

ArgusCore/src/main/java/com/salesforce/dva/argus/service/cache/RedisCacheService.java
```diff
@@ -63,285 +63,293 @@
...
+		for (Map.Entry<String, V> entry : entries.entrySet()) {
+			put(entry.getKey(), entry.getValue(), ttl);
+		}
+	}
+
+	@Override
+	public <V> void expire(String key, int ttl) {
+		try {
+			_jedisClusterClient.expire(key, ttl);
+		} catch (Exception ex) {
```
dilipdevaraj-sfdc: Catch exception here too

ArgusCore/src/main/java/com/salesforce/dva/argus/service/cache/RedisCacheService.java
dilipdevaraj-sfdc: @bsura 
Can you please review

PR #645, opened 2017-09-01
Add new feature to agtable

prestonfff:
Make the table change layout when the original table returns only one row.
Show the first column as tags, the first row as scope+metrics.


cnardi-dev:
Can merge this once updates are made.

PR #649, opened 2017-09-06
Merge 2.8.0 commits to master

xizi-xu:
No files should be changed since recursive strategy is used


xizi-xu:
@prestonfff can you sign the CLA?

prestonfff:
@xizi-xu My account already signed but those commits do not link with email.

PR #651, opened 2017-09-08
React setup

prestonfff:


xizi-xu:
Please do not send any more PR to develop_ui. We are planning on moving the web component of UI into a new repo

PR #653, opened 2017-09-09
Fix annotations related bug.

bsura:
Fix /annotations endpoint returning additional data than the time range queried.


bsura:
@xizi-xu This is the fix for the bug that you reported to me about annotations.

PR #660, opened 2017-09-19
Change from TreeSet to HashSet (Merge into develop)

bsura:


bsura:
@dilipdevaraj-sfdc @rajsarkapally-sfdc Please approve. Its a minor change.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AsyncHbaseSchemaService.java
```diff
@@ -345,7 +346,7 @@ public Object call(ArrayList<ArrayList<KeyValue>> rows) throws Exception {
     	
     	_logger.info("Using FastScan. Will skip rows while scanning.");
     	
-    	final Set<MetricSchemaRecord> records = new TreeSet<>();
+    	final Set<MetricSchemaRecord> records = new HashSet<>();
```
dilipdevaraj-sfdc: What was the original idea behind using TreeSet ?  Do we no longer want the records to be ordered by some comparison method ?

bsura: It was initially a list that got replaced with a set. Ordering is not really important in this case. Hence I changed it to a HashSet.

PR #662, opened 2017-09-19
Fix: getUniqueFastScan returning duplicates (Merge into develop)

bsura:


bsura:
@dilipdevaraj-sfdc @rajsarkapally-sfdc A couple of minor bug fixes. 

Here's some description:
We adding the following check in order to circumvent an OpenTSDB bug where it returns data outside of the requested timespan.

```
if(timestamp > query.getStartTimestamp() && timestamp <= query.getEndTimestamp()) {
     Annotation updated = new Annotation(source, id, type, scope, metric, timestamp);
     updated.setFields(existing.getFields());
     updated.setTags(query.getTags());
     annotations.add(updated);
}
//timestamp is in secs and query.startTimestamp, query.getEndTimestamp could be in milliseconds
```

This led to a regression where no data was returned. This is because OpenTSDB returns sec resolution timestamp for annotations even when explicitly adding ms=true query param in the request. Now, if the query start and end timestamps are in millis and the response annotation timestamps are in seconds, the comparison fails resulting in an empty resultset.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetricSchemaRecord.java
```diff
@@ -120,8 +119,7 @@ public String getScope() {
      *
      * @param  scope  The scope.  Cannot be null or empty.
      */
-    public void setScope(String scope) {
-        SystemAssert.requireArgument(scope != null && !scope.isEmpty(), "Scope cannot be null or empty");
+    public void setScope(String scope) {
```
dilipdevaraj-sfdc: Why have we removed the assert check here. Does this imply scope can be null or empty so defined in Javadoc

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetricSchemaRecord.java
```diff
@@ -139,8 +137,7 @@ public String getMetric() {
      *
      * @param  metric  The metric name.  Cannot be null or empty.
      */
-    public void setMetric(String metric) {
-        SystemAssert.requireArgument(metric != null && !metric.isEmpty(), "Metric cannot be null or empty");
+    public void setMetric(String metric) {
```
dilipdevaraj-sfdc: Same here

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AsyncHbaseSchemaService.java
```diff
@@ -375,8 +375,10 @@ public Object call(ArrayList<ArrayList<KeyValue>> rows) throws Exception {
         	String splits[] = rowKey.split(String.valueOf(ROWKEY_SEPARATOR));
         	String record = (RecordType.METRIC.equals(type) && metadata.tableName.equals(METRIC_SCHEMA_TABLENAME)) || 
         			(RecordType.SCOPE.equals(type) && metadata.tableName.equals(SCOPE_SCHEMA_TABLENAME)) ? splits[0] : splits[1];
-        	
-        	MetricSchemaRecord schemaRecord = _constructMetricSchemaRecord(rowKey, metadata.tableName);
+        
+		MetricSchemaRecord schemaRecord = RecordType.METRIC.equals(type) ?
```
dilipdevaraj-sfdc: Can you provide more context as to how this avoids duplicate records ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AsyncHbaseSchemaService.java
```diff
@@ -375,8 +375,10 @@ public Object call(ArrayList<ArrayList<KeyValue>> rows) throws Exception {
         	String splits[] = rowKey.split(String.valueOf(ROWKEY_SEPARATOR));
         	String record = (RecordType.METRIC.equals(type) && metadata.tableName.equals(METRIC_SCHEMA_TABLENAME)) || 
         			(RecordType.SCOPE.equals(type) && metadata.tableName.equals(SCOPE_SCHEMA_TABLENAME)) ? splits[0] : splits[1];
-        	
-        	MetricSchemaRecord schemaRecord = _constructMetricSchemaRecord(rowKey, metadata.tableName);
+        
+		MetricSchemaRecord schemaRecord = RecordType.METRIC.equals(type) ? 
+				new MetricSchemaRecord(null, record) : new MetricSchemaRecord(record, null);	
+        	//MetricSchemaRecord schemaRecord = _constructMetricSchemaRecord(rowKey, metadata.tableName);
```
dilipdevaraj-sfdc: Remove commented line if unused.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetricSchemaRecord.java
bsura: Yes. It can be null or empty. Thank you for pointing out the JavaDoc inconsistencies.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetricSchemaRecord.java
bsura: Yes. It can be null or empty. Thank you for pointing out the JavaDoc inconsistencies.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/tsdb/DefaultTSDBService.java
```diff
@@ -390,10 +390,25 @@ public void putAnnotations(List<Annotation> annotations) {
                             String type = query.getType();
                             String scope = query.getScope();
                             String metric = query.getMetric();
-                            Long timestamp = existing.getTimestamp();
-                            if(timestamp > query.getStartTimestamp() && timestamp <= query.getEndTimestamp()) {
+			    
+                            //Convert all timestamps to millis, so that we can compare them
+                            long timestamp = existing.getTimestamp();
+                            if(String.valueOf(timestamp).length() < 12) {
```
dilipdevaraj-sfdc: Doesn't it make more sense to check if <=12 to consider it as seconds
Since current time millis  as of today, have 13 digits


https://stackoverflow.com/questions/23929145/how-to-test-if-a-given-time-stamp-is-in-seconds-or-milliseconds
"Can your timestamps be older than 2001? If not i think you're safe to check whether the number has 13 digits or more - if yes you have milliseconds, if not you have seconds.""
Current time
Epoch timestamp: 1505858826
Timestamp in milliseconds: 1505858826000

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AsyncHbaseSchemaService.java
bsura: MetricSchemaRecord schemaRecord = _constructMetricSchemaRecord(rowKey, metadata.tableName);
will construct schemarecord using the rowKey. Row keys can be distinct but since we are only querying for a particular type (scope or metric), they can still be the same. Hence if we construct schemarecord using the entire row key, then in certain cases we will end up duplicate scope/metric.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AsyncHbaseSchemaService.java
bsura: Sure.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/tsdb/DefaultTSDBService.java
bsura: The same post also mentions:

"P.S. you can loosen up the condition to 12 digits or more, if your timestamps can't go back further than 1973. The condition should only be ok up until:

100000000000000 (15 digits) means Wed Nov 16 5138 09:46:40 because this in seconds will have 12 digits and will overlap with your condition"

bsura:
@rajsarkapally-sfdc Can you review this since Dilip is OOO now?

bsura:
@rajsarkapally Please review.

PR #682, opened 2017-10-24
Downsample snapping

rajsarkapally:


rajsarkapally:
@bsura Created pull request just for downsampling

PR #695, opened 2017-11-22
Remove validation of metric objects when deserializing at the WS interface (Merge into develop)

bsura:


bsura:
@dilipdevaraj-sfdc Could you review this, since we chatted about this yesterday?

ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/dto/MetricDto.java
```diff
@@ -189,8 +188,6 @@ public void setUnits(String units) {
      * @param  datapoints  The metric data points.
      */
     public void setDatapoints(Map<Long, Double> datapoints) {
-    	SystemAssert.requireArgument(datapoints != null && !datapoints.isEmpty(), "Datapoints cannot be null or empty.");
```
dilipdevaraj-sfdc: If we remove the validation of datpoints and tag count completely are we now allowing users to submit incorrect metric,and Argus sends 200 OK but TSDB fails.
Shouldn't we still have validation but instead of failing the entire batch of metrics (~5000) remove only the few bad ones ?

bsura: It's handled further down the road, by DefaultTSDBService. So we will be good.

PR #697, opened 2017-12-04
Finding shared alerts, dashboards with filtering by owner

dilipdevaraj-sfdc:


dilipdevaraj-sfdc:
@bsura @rajsarkapally 
Can you please review part 1 of finding shared dashboards alerts.

bsura:
I thought we were implementing a new version of the API. Is that going to come out separately?

dilipdevaraj-sfdc:
@bsura 
Yes that will be a separate commit.
I wanted to send a PR on work done as long as it is complete functional, in spirit of committing small changes and frequently. So we don't have  to do a merge while waiting for completed story.

This is for just
Gets all shared alert/dashboard owned by anyone in Argus system 
/alert/shared 
/alert/shared/meta 
/dashboard/shared 
/dashboard/shared/meta 

Will add other endpoints as discussed once I am done with federation story. 
-----------------
Gets only users alerts/dashboards that can then be filtered by alert/dashboard name and shared status using query params 
v2/alert 
v2/alert/meta 
v2/dashboard 
v2/dashboard/meta 
----------------
Makes sense ?

dilipdevaraj-sfdc:
Sure will do.

dilipdevaraj-sfdc:
@bsura As discussed, this PR is a non breaking change since only new endpoints are added, so it should be good for further code review.

bsura:
Ok I'll take a look.

PR #699, opened 2017-12-04
Fix error when reading annotations - java.text.ParseException: Error …

dilipdevaraj-sfdc:
…parsing Quality value: a decimal numeral '0' or '1' is expected rather than '2'


dilipdevaraj-sfdc:
@rajsarkapally  @bsura  pls review

dilipdevaraj-sfdc:
Quality factor should be within the range 0-1 inclusive

PR #702, opened 2017-12-05
Orchestra Logout exception

rajsarkapally:


ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/resources/AuthResources.java
```diff
@@ -90,8 +90,7 @@ public Response login(@Context UriInfo uriInfo, final CredentialsDto creds) {
     @Description("Terminates a user session. Not needed for existing auth api. This is just a placeholder method.")
     @Path("/logout")
     public Response logout(@Context UriInfo uriInfo) {
-    	Response response = Response.temporaryRedirect(UriBuilder.fromUri(uriInfo.getBaseUri() + "v2/auth/logout").build()).build();
-    	return response;
+    	 return Response.ok("You have logged out.").build();
```
dilipdevaraj-sfdc: After speaking with Raj
Although we have https://github.com/salesforce/ArgusOrchestra/pull/5
To maintain backward compatibility, if users were to call /logout endpoint which is a GET request it does not get redirected to /v2/logout due to issues with VIP. If we are hitting a hostname web server directly then this works, but not with VIP.

PR #706, opened 2017-12-15
 Fix for bugs introduced by serdes of entire alert object while scheduling (Merge into develop)

bsura:


bsura:
@rajsarkapally @dilipdevaraj-sfdc Please review.

bsura:
Ok. I will take a look to see what's causing the tests to fail.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
```diff
@@ -352,6 +352,15 @@ public Alert findAlertByPrimaryKey(BigInteger id) {
 					Map<BigInteger, Map<Metric, Long>> triggerFiredTimesAndMetricsByTrigger = _evaluateTriggers(triggersToEvaluate, 
 							metrics, history);
 					
+					// Update the state of notification objects from the database since the notification contained 
+					// in the serialized alert might be stale. This is because the scheduler only refreshes the alerts
+					// after a specified REFRESH_INTERVAL. And within this interval, the notification state may have changed.
+					// For example, the notification may have been updated to be on cooldown by a previous alert evaluation.
+					// Or it's active/clear status may have changed.
```
dilipdevaraj-sfdc: Here before we evaluate any trigger, we are replacing all in memory notifications with updated notifications from database (thus this will be applicable for all alerts and not just triggered alerts). Although we are doing it in bulk by all notifications ids we may need to check db load pre and post fix.

dilipdevaraj-sfdc: - Can we do this lookup only for alerts that are triggered but before we decide to send out notification.
- Only do it for alerts that have a cron entry lesser than scheduling refresh interval
Eg) less than 15 mins then we check. If greater than 15 mins then it will be automatically taken care by next scheduling period in web server when it reads new state of alert in db before again enqueing.

bsura: 1. No, we need to do this even for alerts that have not triggered. For the case, when an alert needs to be cleared. 
2. This seems to be a legit idea. Let me see how we can achieve this.

bsura: Let's add this as an optimization if the database load increases. Right now, the scheduling interval is determined by each concrete implementation and there's no easy way to determine the refresh interval without harding it into the implementation of DefaultAlertService.

bsura:
@dilipdevaraj-sfdc Please review now.

PR #710, opened 2018-01-22
Change no of tags allowed from 7 to 50.

bsura:


bsura:
@dilipdevaraj-sfdc @rajsarkapally Please review.

PR #716, opened 2018-02-13
ShardedTSDBService and FederatedTSDBService implementation changes

dilipdevaraj-sfdc:
ShardedTSDBService  
- no interpolation performed
- downsampling done at TSDB level

FederatedTSDBService
- no interpolation performed except when avg downsampling used
- substitute user query aggregator with none aggreator , when avg downsampling used


dilipdevaraj-sfdc:
@rajsarkapally 
Can you please review since you had looked at previous review

@bsura 
Can you please take a look too. Although there are two implementations, we will be using ShardedTSDBService

PR #718, opened 2018-02-15
Optimizing the Alert scheduling code to make less queries to the postgres database

sundeepsf:
Optimizing the Alert scheduling code to make less queries to the database during distributed scheduling. The optimizations are two fold -

1. Each scheduling node gets only all the enabled alert ids (instead of the full alert objects)

2. During scheduling, the node gets the actual alert objects that it is responsible for scheduling. Also, during Alert fetch, the code is optimized to fetch all the dependencies of the Alert object instead of having to make subsequent queries like before


bsura:
@sundeepsf You will also have to sign the Salesforce CLA.

PR #722, opened 2018-02-27
BugFix: Metrics without tags are not indexed by ESSchemaService.

bsura:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -53,16 +53,18 @@ public void put(List<Metric> metrics) {
...
         List<Metric> metricsToPut = new ArrayList<>(metrics.size());
+        List<String> keys = new ArrayList<>();
 		
 		for(Metric metric : metrics) {
 			if(metric.getTags().isEmpty()) {
 				String key = _constructTrieKey(metric, null);
 				boolean found = _trie.getValueForExactKey(key) != null;
 		    	if(!found) {
-		    		_trie.putIfAbsent(key, VoidValue.SINGLETON);
+		    		//_trie.putIfAbsent(key, VoidValue.SINGLETON);
```
sundeepsf: no need to keep commented code like this

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -72,7 +74,8 @@ public void put(List<Metric> metrics) {
 					boolean found = _trie.getValueForExactKey(key) != null;
 			    	if(!found) {
 			    		newTags = true;
-			    		_trie.putIfAbsent(key, VoidValue.SINGLETON);
+			    		//_trie.putIfAbsent(key, VoidValue.SINGLETON);
```
sundeepsf: same here

PR #730, opened 2018-03-20
Update moment.js version

xizi-xu:
Update moment.js version that's used by angular-bootstrap-datetimepicker to remove Vulnerability


saaja-sfdc:
@anish FYI

PR #731, opened 2018-03-21
Implementing HBaseAuditService which will store audit data in HBase instead of RDBMS 

sundeepsf:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/audit/HBaseAuditService.java
```diff
@@ -0,0 +1,337 @@
...
+    
+    private static final long PUT_TIMEOUT_MS = 30 * 1000;
+    private static final long SCAN_TIMEOUT_MS = 60 * 1000;
+    private static final long SHUTDOWN_TIMEOUT_MS = 30 * 1000;
+	
+    
+    //~ Instance fields ******************************************************************************************************************************
+    
+	@SLF4JTypeListener.InjectLogger
+    private Logger _logger;
```
bsura: Avoid injecting the logger. Sometimes it's become difficult to test. Instead use something like:
private Logger _logger = LoggerFactory.getLogger(getClass());

PR #732, opened 2018-03-29
ESSchemaService enhancements

bsura:
- Add capability to configure XXHASH algorithm to generate document ids
- Better error/exception handling and removal of records from trie
- Disable writes to TRIE when old gen size is approaching JVM heap capacity
- Improve latency for queries involving terms aggregation by disabling building of global ordinals on search requests
- Instrument indexing/query latencies


ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -167,5 +214,62 @@ public String getDefaultValue() {
...
+		private void _checkOldGenUsage() {
+			List<MemoryPoolMXBean> memoryPoolBeans = ManagementFactory.getMemoryPoolMXBeans();
+			for (MemoryPoolMXBean bean : memoryPoolBeans) {
+				if (bean.getType() == MemoryType.HEAP) {
+					String name = bean.getName().toLowerCase();
+					if (name.contains("old") || name.contains("tenured")) {
+						long oldGenUsed = bean.getUsage().getUsed();
+						_logger.info("Old Gen Memory = {} bytes", oldGenUsed);
+						_logger.info("Max JVM Memory = {} bytes", MAX_MEMORY);
+						if (oldGenUsed > 0.90 * MAX_MEMORY) {
```
sundeepsf: not sure if this will work always. Heap memory goes to young gen also and if that is 10 percent, then this will never be true. Can't we just go with total heap memory used instead of old gen memory?

bsura: Might not work for smaller heaps and cause FULL GCs, but for larger heaps (say something like 16G) eden space of 1.6G should be more than enough.

sundeepsf: eden space AFAIK is set using newRatio or newSizePercent. Not sure what these values are in our current setup

sundeepsf: In any case, comparing the current old gen memory to max configured old gen memory will be safer

bsura: We are using G1GC in our current setup which doesn't require us to set young to old ratio. This is only a stop-gap best effort approach to avoid from running into FULL GCs and eventually OOME.

sundeepsf: G1 uses -XX:G1NewSizePercent and -XX:G1MaxNewSizePercent. The defaults for those are 5 and 60 according to this - http://www.oracle.com/technetwork/articles/java/g1gc-1984535.html

So, if new gen is taking more than 10 percent of the heap, the condition will never be true even if we are running OOM

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/MetricSchemaRecordList.java
```diff
@@ -61,18 +92,20 @@ public void serialize(MetricSchemaRecordList list, JsonGenerator jgen, Serialize
...
+				jgen.writeRaw(mapper.writeValueAsString(entry.getValue()));
 				jgen.writeRaw(System.lineSeparator());
 			}
+			
 		}
     	
     }
 	
+	
 	static class Deserializer extends JsonDeserializer<MetricSchemaRecordList> {
```
sundeepsf: should this change also as the serialization mechanism has changed?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
bsura: Hmm. In that case, it would be beneficial to compare against max old gen available.

sundeepsf: It sounds like with G1, it dynamically adjusts the max heap available to old gen and young gen. Its not a fixed number like with the other GC algorithms

PR #737, opened 2018-04-21
Add qurey window and no. expanded time series as attributes from metric reader back to perffilter

dilipdevaraj-sfdc:
- Modified approach since we don't have to make additional call to expand time series again for measurement.


dilipdevaraj-sfdc:
@rajsarkapally  can you please review the modified approach.

PR #738, opened 2018-04-24
Add GROUPBYTAG function

melan:
Semantic of the function is a mix of ALIASBYTAG and GROUPBY:
`GROUPBYTAG(some:metric, #tag1#, #tag2#, #SUM#, #union#)`

The 1st group after the metric query is a list of tags. The group should have 1 or more elements.
The 2nd group - is a name of a reduce function, the code checks the name against `TransformFactory.Function`, the check is case sensitive.
The 3rd group are transformation constants applicable to the reduce function.

The code throws an error when where are no tags or when the function wasn't found.


melan:
@cnardi-dev I signed the CLA, is there anything else I should do?

cnardi-dev:
Looks good, thnx for contributing!

PR #739, opened 2018-04-25
Made changes to detect metric data lag and disable alert evaluation if there is a data lag.

sundeepsf:
Added several config params to be able to specify which expression to use for detecting data lag, threshold to use for the same etc.


ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DefaultMonitorService.java
```diff
@@ -84,604 +86,621 @@
...
-            }
-        }
-    }
+
+			_monitorThread.start();
+			_logger.info("System monitor thread started.");
+
+			if (Boolean.valueOf(_sysConfig.getValue(com.salesforce.dva.argus.system.SystemConfiguration.Property.DATA_LAG_MONITOR_ENABLED))) {
+				_dataLagMonitorThread = new DataLagMonitor(_sysConfig, _metricService, _mailService);
+				_dataLagMonitorThread.start();
```
dilipdevaraj-sfdc: Do we plan to stop the data lag monitor thread once it has started ? 
similar to monitor thread stopRecordingCounters

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DataLagMonitor.java
```diff
@@ -0,0 +1,120 @@
...
+						}
+						continue;
+					}
+				}
+				if(isDataLagging) {
+					isDataLagging = false;
+					sendDataLagEmailNotification();
+				}
+			}catch(Exception e) {
+				_logger.error("Data lag monitor thread sleep interrupted - " + e.getMessage());
```
dilipdevaraj-sfdc: When will it enter this state if we are not interrupting this thread ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
```diff
@@ -355,6 +355,20 @@ public void updateNotificationsActiveStatusAndCooldown(List<Notification> notifi
 			String logMessage = null;
 			History history = new History(addDateToMessage(JobStatus.STARTED.getDescription()), SystemConfiguration.getHostname(), alert.getId(), JobStatus.STARTED);
 			
+			if(Boolean.valueOf(_configuration.getValue(com.salesforce.dva.argus.system.SystemConfiguration.Property.DATA_LAG_MONITOR_ENABLED))){
+				if(_monitorService.isDataLagging()) {
+					logMessage = MessageFormat.format("Skipping evaluating the alert with id: {0}. because metric data was lagging", alert.getId());
+					_logger.info(logMessage);
+					_appendMessageNUpdateHistory(history, logMessage, null, 0);
+					history = _historyService.createHistory(alert, history.getMessage(), history.getJobStatus(), history.getExecutionTime());
```
dilipdevaraj-sfdc: In addition to history mesage, is there a way for user to know  from history alert evaluation was SKIPPED
similar to JobStatus - STARTED, FAILURE

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DataLagMonitor.java
```diff
@@ -0,0 +1,120 @@
...
+	private MailService _mailService;
+
+	private static final Long SLEEP_INTERVAL_MILLIS = 60*1000L;
+
+	private final Logger _logger = LoggerFactory.getLogger(DataLagMonitor.class);
+
+	public DataLagMonitor(SystemConfiguration sysConfig, MetricService metricService, MailService mailService) {
+		_metricService = metricService;
+		_mailService = mailService;
+		_dataLagQueryExpression = sysConfig.getValue(com.salesforce.dva.argus.system.SystemConfiguration.Property.DATA_LAG_QUERY_EXPRESSION);
```
dilipdevaraj-sfdc: In addition to checking if golden metric is present, since we round robin to Kafka, we should evaluate if Kafka lag measurement on a partition too is needed. 
If lag increases on 1 partition we should still be notified even though we are tracking by "golden metric" SUM UNION


To validate, commit client connected to 1 tsdb daemon
-  bring down tsdb daemon
- bring down commit client

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DefaultMonitorService.java
sundeepsf: Good point, i missed changing that method. Will fix it

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DataLagMonitor.java
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	@Override
+	public void run() {
+		while (!isInterrupted()) {
+			try {
+				sleep(SLEEP_INTERVAL_MILLIS);
+				long currTime = System.currentTimeMillis();
+				List<Metric> metrics = _metricService.getMetrics(_dataLagQueryExpression, currTime);
+				if(metrics==null || metrics.isEmpty()) {
```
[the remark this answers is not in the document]
sundeepsf: Its very small logic, felt its overkill to refactor

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DataLagMonitor.java
sundeepsf: Will fix it

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
sundeepsf: Yes, thats good idea. Will add that to history

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DataLagMonitor.java
sundeepsf: I felt its much more foolproof to check the actual end data last inserted timestamp, instead of relying on kafka lag. Even if there is kafka lag, but end users are not affected, then we should evaluate the alerts. I agree this may not catch all cases(like a single dc being down or single tsdb daemon down), but its kind of best effort for now

PR #743, opened 2018-05-08
1)Add ts field as date type 2)change POST ES request to use index fro…

dilipdevaraj-sfdc:
…m create


dilipdevaraj-sfdc:
@sundeepsf Can you review.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -939,11 +939,13 @@ private ObjectNode _createMappingsNode() {
...
-    	propertiesNode.put(RecordType.TAGK.getName(), _createFieldNode());
-    	propertiesNode.put(RecordType.TAGV.getName(), _createFieldNode());
-    	propertiesNode.put(RecordType.NAMESPACE.getName(), _createFieldNode());
+    	propertiesNode.put(RecordType.SCOPE.getName(), _createFieldNode("text"));
+    	propertiesNode.put(RecordType.METRIC.getName(), _createFieldNode("text"));
+    	propertiesNode.put(RecordType.TAGK.getName(), _createFieldNode("text"));
+    	propertiesNode.put(RecordType.TAGV.getName(), _createFieldNode("text"));
+    	propertiesNode.put(RecordType.NAMESPACE.getName(), _createFieldNode("text"));
+    	
+    	propertiesNode.put("mts", _createFieldNode("date"));
```
sundeepsf: can this be a better name like createdDateMillis etc? Also better to add a reference to RecordType enum like above fields

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -939,11 +939,13 @@ private ObjectNode _createMappingsNode() {
     	ObjectMapper mapper = new ObjectMapper();
     	
     	ObjectNode propertiesNode = mapper.createObjectNode();
-    	propertiesNode.put(RecordType.SCOPE.getName(), _createFieldNode());
-    	propertiesNode.put(RecordType.METRIC.getName(), _createFieldNode());
-    	propertiesNode.put(RecordType.TAGK.getName(), _createFieldNode());
-    	propertiesNode.put(RecordType.TAGV.getName(), _createFieldNode());
-    	propertiesNode.put(RecordType.NAMESPACE.getName(), _createFieldNode());
+    	propertiesNode.put(RecordType.SCOPE.getName(), _createFieldNode("text"));
```
sundeepsf: maybe this text can be a constant?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
dilipdevaraj-sfdc: It is mts to shorten payload - modified time stamp. 
MetricSchemaRecord should be agnostic of timestamp field since the essential field - scope, metric, tag are what drives MetricSchemaRecord
When the hash gets computed it should not use the newer mts field
hence I added it as part of serialization at end only similar to how we are using _id field

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
dilipdevaraj-sfdc: We are using it only in this method so it should be fine.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
sundeepsf: is mts used in ES document? It maybe ok to just used for message transport, but it looks kind of ugly to use this as db field

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
dilipdevaraj-sfdc: I have made this change.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
dilipdevaraj-sfdc: mts is a field in ES document. It is already an existing field on our ES cluster after we created it via REST operation.

sundeepsf: ok :(

PR #745, opened 2018-05-14
Fix conflicting logger implementations error

colbyguan:
This should fix the `java.lang.ClassCastException: org.slf4j.impl.Log4jLoggerAdapter cannot be cast to ch.qos.logback.classic.Logger` error that happens when you run webservices locally (documented as a common error in the internal "Setting Up Argus Dev" document)

### DIFF
--- a/ArgusCore/pom.xml
+++ b/ArgusCore/pom.xml
@@ -159,6 +159,14 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


colbyguan:
@rajsarkapally I have confirmed locally that logging works as normal. For unit tests, I assume only logs of level ERROR are supposed to output; which they do.

PR #749, opened 2018-05-19
Add GPG key generation / publish to finalize automated deploy

justinharringa:
This change adds a script to generate a new GPG key and use it for project publishing. This should take care of what was left to get this automatically deployed. 

NOTE: I also bumped the `project.version` file to 2.12.0 since that's what the pom.xml is on. You'll need to remember to bump that in order to properly set your versions. An alternative approach would be to keep your project version in your `pom.xml` set to a snapshot version and then let the deployment handle that. But that's up to you.

The following keys have been set in the Travis CI project to support this script.
GPG_EMAIL=argus-no-reply@salesforce.com
GPG_NAME=argus release


justinharringa:
PS: Props to @jamesward for his work on this alternative approach that doesn't require storing more secrets.

justinharringa:
@dilipdevaraj-sfdc don't forget to get these changes into `master` so that it can automatically deploy. :)

dilipdevaraj-sfdc:
@justinharringa 

We have merged the changes to master and set the correct project version 2.15.0 at https://github.com/salesforce/Argus/blob/master/project.version

However the Travis CI that is trying to publish this from master branch fails. It is waiting for a long time for project.version to be set then timing out and failing. See
-----------------------
https://travis-ci.org/salesforce/Argus/builds/394320897
Enter the new version to set ${project.version}: :

justinharringa:
Drat... Travis isn't able to source a couple of files. I sent the following PR to `master` to address that: https://github.com/salesforce/Argus/pull/763

PR #758, opened 2018-06-18
1)Add num time series from tsdb for a user 2)Add num results from dis…

dilipdevaraj-sfdc:
…covery service 3)Add num queries to discovery service


dilipdevaraj-sfdc:
@nkunal 
pls review

PR #763, opened 2018-06-19
Source scripts from absolute path (${TRAVIS_BUILD_DIR}/script)

justinharringa:
@dilipdevaraj-sfdc adding absolute path to scripts / files

### DIFF
--- a/.travis.deploy.sh
+++ b/.travis.deploy.sh
@@ -1,14 +1,15 @@
 #!/usr/bin/env bash
 # Generate & push code signing keys to keyserver.ubuntu.com, source into environment
-source .travis.gpg.sh
+source ${TRAVIS_BUILD_DIR}/.travis.gpg.sh
 
 # Utilize tokenized `.travis.settings.xml` to do deployment
-cp .travis.settings.xml $HOME/.m2/settings.xml
+cp ${TRAVIS_BUILD_DIR}/.travis.settings.xml ${HOME}/.m2/settings.xml
 
 # PROJECT_VERSION in the commands below should be incremented and sourced from project.version
-source project.version
+source ${TRAVIS_BUILD_DIR}/project.version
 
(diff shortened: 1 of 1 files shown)


dilipdevaraj-sfdc:
@justinharringa 
Auto deploy is still failing after this change. Do you know what may need to be modified?

My thinking is Argus/pom.xml has ${project.version} parameter defined 
See https://github.com/salesforce/Argus/blob/master/pom.xml 
Value is set in two places. 
1) <project.version>2.15.0</project.version>
2) https://github.com/salesforce/Argus/blob/master/project.version

The build of individual artifacts get correctly generated with 2.15.0 however only while publishing to maven it is still failing ?

-----------------------
https://travis-ci.org/salesforce/Argus/builds/394347552
Looking through logs I see
[WARNING] Some problems were encountered while building the effective model for com.salesforce.argus:argus-client:jar:2.15.0
[WARNING] 'version' contains an expression but should be a constant. @ com.salesforce.argus:argus:${project.version}, /home/travis/build/salesforce/Argus/pom.xml, line 7, column 14

and then it finally stops and fails at 
Enter the new version to set ${project.version}: :

justinharringa:
Yeah, the main issue is that the source command in .travis.deploy.sh for [GPG](https://github.com/salesforce/Argus/blob/master/.travis.deploy.sh#L3) and [project.version](https://github.com/salesforce/Argus/blob/master/.travis.deploy.sh#L9) is not working. So when we do [mvn versions:set](https://github.com/salesforce/Argus/blob/master/.travis.deploy.sh#L13) it runs with no value.

dilipdevaraj-sfdc:
Ok. Then at this time auto deploy to maven would not work.
Do you know how much effort is needed for setting GPG and project.version correctly so deploy to maven works.

In the interim can you please submit an additional PR from your "absolute-path-for-deploy-scripts" branch to Argus develop branch. 
This is so that we have master and develop in sync. (Note - develop is our running incremental branch)

PR #792, opened 2018-06-25
Log discovery queries that is above a static configurable threshold

dilipdevaraj-sfdc:


dilipdevaraj-sfdc:
@nkunal 
Can you pls review

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/CachedDiscoveryService.java
```diff
@@ -129,10 +134,50 @@ private void _disposeExecutorService() {
 			queries.add(query);
 		}
 		
-		_logger.debug("Time to get matching queries in ms: " + (System.nanoTime() - start) / 1000000);
+		long timeToGetQueriesMillis = (System.nanoTime() - start) / 1000000;
```
nkunal: Do we want to log the es query time separately than what was received from cache ? perhaps , we could move the timeToGetQueriesMillis in the cache miss block and log it there ?

dilipdevaraj-sfdc: I would tend to think this is the net time taken before the customer sees the query being returned independent of whether it is from cache or from underlying discovery service (ES or Hbase or any other implementation). Hence I logged it here.

dilipdevaraj-sfdc: Moved the code to cache miss block.

PR #802, opened 2018-07-06
Redis incrby

pratiksha-shah:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/cache/RedisCacheService.java
```diff
@@ -125,6 +125,18 @@ public RedisCacheService(SystemConfiguration config) {
 		return map;
 	}
 
+	@Override
+	public <V> Long incrby(String key, Long value) {
```
dilipdevaraj-sfdc: pls rename to camelcase incrBy

ArgusCore/src/main/java/com/salesforce/dva/argus/service/cache/RedisCacheService.java
```diff
@@ -125,6 +125,18 @@ public RedisCacheService(SystemConfiguration config) {
 		return map;
 	}
 
+	@Override
```
dilipdevaraj-sfdc: Pls add this corresponding method to interface CacheService.java if you are using @Override annotation

dilipdevaraj-sfdc: Also add the java doc of what this method would do - like it would increment and return the incremented value, atomically.
Also add a note that returnValue need not be currentValue + (incrementedByValue) for a client, if this operation is performed parallelly by other clients.

PR #812, opened 2018-07-12
Add new scopenames index with scope and mts field

dilipdevaraj-sfdc:
Have separate bloom filter for scope names only
Remove async put from ElasticSchema Service. ES writes will be synchronous only
Make number of shards, replication factor, scope index name, scope type name configurable

### DIFF
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeOnlySchemaRecord.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeOnlySchemaRecord.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright (c) 2016, Salesforce.com, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of Salesforce.com nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.salesforce.dva.argus.entity;
+
+import java.text.MessageFormat;
+
+/**
+ * Represents a search result row for scope names discovery queries.
+ *
+ * @author  Dilip Devaraj (ddevaraj@salesforce.com)
+ */
+public class ScopeOnlySchemaRecord {
+
+	//~ Instance fields ******************************************************************************************************************************
+
+	private String scope;
+
+	//~ Constructors *********************************************************************************************************************************
+
+	/** Creates a new ScopeOnlySchemaRecord object. */
+	public ScopeOnlySchemaRecord() { }
+
+
+	/**
+	 * Creates a new ScopeOnlySchemaRecord object.
+	 *
+	 * @param  scope      The metric schema scope.
+	 */
+	public ScopeOnlySchemaRecord(String scope) {
+		setScope(scope);
+	}
+
+	/**
+	 * Indicates the scope associated with the result.
+	 *
+	 * @return  The scope.  Can be null or empty.
+	 */
+	public String getScope() {
+		return scope;
+	}
+
+	/**
+	 * Specifies the scope associated with the result.
+	 *
+	 * @param  scope  The scope.  Can be null or empty.
+	 */
+	public void setScope(String scope) { 
+		this.scope = scope;
+	}
+
+
+	@Override
+	public int hashCode() {
+		final int prime = 31;
+		int result = 1;
+
+		result = prime * result + ((scope == null) ? 0 : scope.hashCode());
+		return result;
+	}
+
+	@Override
+	public boolean equals(Object obj) {
+		if (this == obj) {
+			return true;
+		}
+		if (obj == null) {
+			return false;
+		}
+		if (getClass() != obj.getClass()) {
+			return false;
+		}
+
+		ScopeOnlySchemaRecord other = (ScopeOnlySchemaRecord) obj;
+
+		if (scope == null) {
+			if (other.scope != null) {
+				return false;
+			}
+		} else if (!scope.equals(other.scope)) {
+			return false;
+		}
+		return true;
+	}
+
+	@Override
+	public String toString() {
+		return MessageFormat.format("ScopeOnlySchemaRecord = (Scope = {0}", scope);
+	}
+
+	public static String print(ScopeOnlySchemaRecord msr) {
+		StringBuilder sb = new StringBuilder(msr.getScope());
+		return sb.toString();
+	}
+}
+/* Copyright (c) 2016, Salesforce.com, Inc.  All rights reserved. */
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
@@ -27,15 +27,28 @@ import com.salesforce.dva.argus.service.SchemaService;
 import com.salesforce.dva.argus.system.SystemAssert;
 import com.salesforce.dva.argus.system.SystemConfiguration;
 
+/**
+ * Implementation of the abstract schema service class
+ *
+ * @author  Dilip Devaraj (ddevaraj@salesforce.com)
+ */
 public abstract class AbstractSchemaService extends DefaultService implements SchemaService {
 	private static final long POLL_INTERVAL_MS = 10 * 60 * 1000L;
 	private static final int DAY_IN_SECONDS = 24 * 60 * 60;
 	private static final int HOUR_IN_SECONDS = 60 * 60;
+
+	/* Have two separate bloom filters one for metrics schema and another for scope names schema.
+	 * Since scopes will continue to repeat more often on subsequent kafka batch reads, we can easily check this from the  bloom filter for scopes only. 
+	 * Hence we can avoid the extra call to populate scopenames index on ES in subsequent Kafka reads.
+	 */
 	protected static BloomFilter<CharSequence> bloomFilter;
+	protected static BloomFilter<CharSequence> bloomFilterScopeOnly;
 	private Random rand = new Random();
 	private int randomNumber = rand.nextInt();
 	private int bloomFilterExpectedNumberInsertions;
 	private double bloomFilterErrorRate;
+	private int bloomFilterScopeOnlyExpectedNumberInsertions;
+	private double bloomFilterScopeOnlyErrorRate;	
 	private final Logger _logger = LoggerFactory.getLogger(getClass());
 	private final Thread _bloomFilterMonitorThread;
 	protected final boolean _syncPut;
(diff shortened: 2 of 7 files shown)


dilipdevaraj-sfdc:
@nkunal  pls review

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -598,71 +648,79 @@ private void _upsert(List<MetricSchemaRecord> records) {
...
 
-		ResponseListener responseListener = new ResponseListener() {
+		try {
+			PutResponse putResponse = new ObjectMapper().readValue(strResponse, PutResponse.class);
+			//TODO: If response contains HTTP 429 Too Many Requests (EsRejectedExecutionException), then retry with exponential back-off.
+			if(putResponse.errors) {
+				List<ScopeOnlySchemaRecord> recordsToRemove = new ArrayList<>();
+				for(Item item : putResponse.items) {
+					if(item.create != null && item.create.status != HttpStatus.SC_CONFLICT && item.create.status != HttpStatus.SC_CREATED) {
+						_logger.warn("Failed to index metric. Reason: " + new ObjectMapper().writeValueAsString(item.create.error));
```
nkunal: log msg should mention "scope" instead of metric

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -673,6 +731,14 @@ protected void _addToBloomFilter(List<MetricSchemaRecord> records){
 		}		
 	}
 
+	protected void _addToBloomFilterScopeOnly(List<ScopeOnlySchemaRecord> records){
+		_logger.info("Adding {} records into scope only bloom filter.", records.size());
+		for(ScopeOnlySchemaRecord record : records) {		
+			String key = constructScopeOnlyKey(record.getScope());		
+			bloomFilter.put(key);
```
nkunal: this should be bloomFilterScopeOnly

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -952,17 +1027,32 @@ private ObjectNode _createMappingsNode() {
...
 	}
 
+	private ObjectNode _createScopeMappingsNode() {
+		ObjectMapper mapper = new ObjectMapper();
+
+		ObjectNode propertiesNode = mapper.createObjectNode();
+		propertiesNode.put(RecordType.SCOPE.getName(), _createFieldNode(FIELD_TYPE_TEXT));
+
+		propertiesNode.put("mts", _createFieldNodeNoAnalyzer(FIELD_TYPE_DATE));
+
```
nkunal: Since we are creating a fresh new index, we might as well add the "cts" field, so then later on we don't have to worry about updating the mappings via rest call.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -598,71 +648,79 @@ private void _upsert(List<MetricSchemaRecord> records) {
...
+				.append(SCOPE_TYPE_NAME)				
 				.append("/")
 				.append("_bulk")
 				.toString();
 
-		MetricSchemaRecordList msrList = new MetricSchemaRecordList(records, _idgenHashAlgo);
-		StringEntity entity;
+		String strResponse = "";
+
+		ScopeOnlySchemaRecordList scopeOnlySchemaRecordList = new ScopeOnlySchemaRecordList(records, _idgenHashAlgo);
```
nkunal: Since this is new index, should we start using xxHash instead of MD5 ? The old index probably can't be switched easily but at least the new indexes can start using xxHash

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -215,6 +269,8 @@ protected int getNumHoursUntilTargetHour(int targetHour){
 		SYNC_PUT("service.property.schema.sync.put", "false"),
 		BLOOMFILTER_EXPECTED_NUMBER_INSERTIONS("service.property.schema.bloomfilter.expected.number.insertions", "40"),
 		BLOOMFILTER_ERROR_RATE("service.property.schema.bloomfilter.error.rate", "0.00001"),
+		BLOOMFILTER_SCOPE_ONLY_EXPECTED_NUMBER_INSERTIONS("service.property.schema.bloomfilter.scope.only.expected.number.insertions", "40"),
```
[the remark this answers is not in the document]
dilipdevaraj-sfdc: 1000000
just in case we have a burst of scopes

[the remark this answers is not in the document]
nkunal: sounds good

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
dilipdevaraj-sfdc: done

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
dilipdevaraj-sfdc: done

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
dilipdevaraj-sfdc: Agreed. https://kb.pomfort.com/silverstack/offloadandbackup/checksum-verification-process/
This is a runtime property configuration so we can choose XXHASH

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
dilipdevaraj-sfdc: Agreed. https://kb.pomfort.com/silverstack/offloadandbackup/checksum-verification-process/
This is a runtime property configuration so we can choose XXHASH

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
dilipdevaraj-sfdc: done.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
nkunal: the only issue is this property is shared across both indexes, unless we separate the configs for both.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -181,6 +215,26 @@ protected String constructKey(String scope, String metric, String tagk, String t
 		return sb.toString();
 	}
 
+	protected String constructScopeOnlyKey(Metric metric) {
```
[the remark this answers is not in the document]
dilipdevaraj-sfdc: done

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
dilipdevaraj-sfdc: I feel that having replication factor, shard count be configurable is fine since they can vary across different index.
But having a separate hash algorithm for two indexes is not logical since we want the hash algorithm that we choose to be uniform across all indexes since we would have determined that this is the best hashing for our given data independent of number of indexes.
So I am thinking we can continue to use MD5 so its consistent, unless we see a huge lag buildup caused by writing to multiple indexes. 
Let me know what you think ?

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeOnlySchemaRecord.java
```diff
@@ -0,0 +1,124 @@
...
+	 *
+	 * @param  scope  The scope.  Can be null or empty.
+	 */
+	public void setScope(String scope) { 
+		this.scope = scope;
+	}
+
+
+	@Override
+	public int hashCode() {
```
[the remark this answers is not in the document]
dilipdevaraj-sfdc: Not directly as of now. But generally done, so when we are using this in a hash key this method will be invoked.
(Most Argus entities @Overrides the equals and hashCode method)

dilipdevaraj-sfdc:
@nkunal 
Addressed comments

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeOnlySchemaRecord.java
```diff
@@ -0,0 +1,124 @@
...
+		if (obj == null) {
+			return false;
+		}
+		if (getClass() != obj.getClass()) {
+			return false;
+		}
+
+		ScopeOnlySchemaRecord other = (ScopeOnlySchemaRecord) obj;
+
+		if (scope == null) {
```
naveenreddykarri: this can be simplified as return Object.equals(scope, other.scope);

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeOnlySchemaRecord.java
```diff
@@ -0,0 +1,124 @@
...
+		}
+		return true;
+	}
+
+	@Override
+	public String toString() {
+		return MessageFormat.format("ScopeOnlySchemaRecord = (Scope = {0}", scope);
+	}
+
+	public static String print(ScopeOnlySchemaRecord msr) {
```
naveenreddykarri: This isn't printing but returning a toString of this class. Should we merge this method and toString() method?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ScopeOnlySchemaRecordList.java
```diff
@@ -0,0 +1,154 @@
...
+	}
+	
+	ScopeOnlySchemaRecord getRecord(String id) {
+		return _idToSchemaRecordMap.get(id);
+	}
+	
+	static class Serializer extends JsonSerializer<ScopeOnlySchemaRecordList> {
+
+		@Override
+		public void serialize(ScopeOnlySchemaRecordList list, JsonGenerator jgen, SerializerProvider provider)
```
naveenreddykarri: I see that SerializerProvider provider is not being used. Is that expected.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -248,6 +280,26 @@ protected void implementationSpecificPut(List<Metric> metrics) {
...
+		for(String scope : scopeNames) {
+			records.add(new ScopeOnlySchemaRecord(scope));
+
+			if(records.size() == _bulkIndexingSize) {
+				fracturedList.add(records);
+				records = new ArrayList<>(_bulkIndexingSize);
+			}
+		}
+
+		fracturedList.add(records);
```
naveenreddykarri: should this be if (records.size() > 0) {
    fracturedList.add(records);
}

naveenreddykarri: or if (records.isEmpty())

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -598,71 +648,79 @@ private void _upsert(List<MetricSchemaRecord> records) {
...
-				//TODO: Retry with exponential back-off for handling EsRejectedExecutionException/RemoteTransportException/TimeoutException??
-				_logger.warn("Failed to execute the indexing request.", e);
-			}
-		};
+		} catch(IOException e) {
+			throw new SystemException("Failed to parse reponse of put metrics. The response was: " + strResponse, e);
+		}
+
+		try {
+			PutResponse putResponse = new ObjectMapper().readValue(strResponse, PutResponse.class);
```
naveenreddykarri: isnt this repeating what we have in line 679?

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeOnlySchemaRecord.java
dilipdevaraj-sfdc: current line looks fine.
It is a standard way of comparing object instance equality. In this case it happens to compare String

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeOnlySchemaRecord.java
dilipdevaraj-sfdc: fixed

dilipdevaraj-sfdc: I have removed this method.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
dilipdevaraj-sfdc: It works fine. 
List.add allows an empty element list to be added.
records will never be null so there is no explicit null check needed

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
dilipdevaraj-sfdc: fixed

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeOnlySchemaRecord.java
naveenreddykarri: meant instead of 
		if (scope == null) {
			if (other.scope != null) {
				return false;
			}
		} else if (!scope.equals(other.scope)) {
			return false;
		}
		return true;

we can just say return Objects.equals(scope, other.scope);

PR #813, opened 2018-07-13
Sync TSD write-endpoints iterable and customizable logback output file names

colbyguan:
These changes are necessary for our new consumer to work

When unsynchronized, with two threads both calling _roundRobinIterator.next(), one has a chance to move the iterator offset past the end, without the other cycling the iterator first, resulting in a `NoSuchElementException` 

Logback change: eg. If two different ArgusCore-dependent Java processes are running, they will both try to log to `argus.log`, this change allows setting the filename to `argus_process1.log`, etc., from a command line argument


colbyguan:
@dilipdevaraj-sfdc @sundeepsf

ArgusCore/src/main/java/com/salesforce/dva/argus/service/tsdb/AbstractTSDBService.java
```diff
@@ -303,7 +303,10 @@ public void putMetrics(List<Metric> metrics) {
 		requireArgument(TSDB_DATAPOINTS_WRITE_MAX_SIZE > 0, "Max Chunk size can not be less than 1");
 		requireArgument(metrics != null, "Metrics can not be null");
 
-		String endpoint = _roundRobinIterator.next();
+		String endpoint;
+		synchronized (_roundRobinIterator) {
+			endpoint = _roundRobinIterator.next();
+		}
```
nkunal: @colbyguan _roundRobinIterator is still accessed unprotected down below. Do you think protection is needed everywhere ?
If we are just trying to cycle through endpoints, a better approach might be to just use an AtomicInteger and use it access the array of endpoints, and then reset it to 0 position once gone through all endpoints.

nkunal: at some point we should replace this code, all of the synchronization is not needed to just get a endpoint from a list of endpoints, wasted resource. For now you can proceed with your commit.

colbyguan:
@nkunal I don't know if there is a way to ensure item-ordering without synchronizing on the whole `queue.add(iter.next())` call; instead I can test that each `endpoint` element is added an even amount to prove that the iterator is doing a fair round robin (most recent commit)

colbyguan:
@dilipdevaraj-sfdc updated

ArgusCore/src/main/java/com/salesforce/dva/argus/service/tsdb/AbstractTSDBService.java
```diff
@@ -193,7 +196,7 @@ public AbstractTSDBService(SystemConfiguration config, MonitorService monitorSer
 
 			_writeHttpClient = getClient(connCount / 2, connTimeout, socketTimeout,tsdbConnectionReuseCount, _writeEndpoints);
 
-			_roundRobinIterator = Iterables.cycle(_writeEndpoints).iterator();
+			_roundRobinIterator = constructCyclingIterator(_writeEndpoints);
```
dilipdevaraj-sfdc: @colbyguan  @nkunal 
Approving PR
Sharing some notes
- So we are having our own constructCyclingIterator since google guava does not support a concurrent iterable. 
- A suitable library that already supports concurrent iterable was not found.
- This problem of concurrency surfaces when we have a client that does a lot of frequent writes from different threads with very less wait time between each write.

PR #816, opened 2018-07-16
Alert evaluation improvements

sundeepsf:
1. Adding support for getting regular GOC/email notifications in case of missing data

2. Add new alert evaluation kpi metric which tracks how long does it take for an alert to trigger based on an incoming metric

3. Making the whitelisted scopes match the regex instead of substring match

4. Other minor changes and bug fixes


ArgusCore/src/main/java/com/salesforce/dva/argus/util/AlertUtils.java
```diff
@@ -0,0 +1,16 @@
+package com.salesforce.dva.argus.util;
+
+import java.util.Set;
+import java.util.regex.Pattern;
+
+public class AlertUtils {
+
+	public static boolean isScopePresentInWhiteList(String expression, Set<String> whiteListedScopeRegexes) {
+		for(String scopeRegex : whiteListedScopeRegexes) {
+			if(Pattern.compile(":"+scopeRegex+":").matcher(expression.toLowerCase()).find()) {
```
rajsarkapally: Since you are converting the expression to lower case, is it a good idea to convert the scopeRegex to lowercase?

ArgusCore/src/main/java/com/salesforce/dva/argus/util/AlertUtils.java
```diff
@@ -0,0 +1,16 @@
+package com.salesforce.dva.argus.util;
+
+import java.util.Set;
+import java.util.regex.Pattern;
+
+public class AlertUtils {
+
+	public static boolean isScopePresentInWhiteList(String expression, Set<String> whiteListedScopeRegexes) {
```
rajsarkapally: Since it is a public method, provide method doc

ArgusCore/src/main/java/com/salesforce/dva/argus/util/AlertUtils.java
sundeepsf: Good point. I did some more refactoring on top to store the patterns in lowercase and the compare them

ArgusCore/src/main/java/com/salesforce/dva/argus/util/AlertUtils.java
```diff
@@ -0,0 +1,16 @@
+package com.salesforce.dva.argus.util;
+
+import java.util.List;
+import java.util.regex.Pattern;
+
+public class AlertUtils {
+
+	public static boolean isScopePresentInWhiteList(String expression, List<Pattern> regexPatterns) {
+		for(Pattern regexPattern : regexPatterns) {
```
dilipdevaraj-sfdc: are we planning to restrict this just for regex scopes ?
If you specify a scope say "kafka.broker.*.ajna_local" and some user's metrics has this in metric name field wouldn't it match in this case too ?

ArgusCore/src/main/java/com/salesforce/dva/argus/util/AlertUtils.java
```diff
@@ -0,0 +1,16 @@
+package com.salesforce.dva.argus.util;
+
+import java.util.List;
+import java.util.regex.Pattern;
+
+public class AlertUtils {
```
dilipdevaraj-sfdc: Also specify javadoc for class

ArgusCore/src/main/java/com/salesforce/dva/argus/util/AlertUtils.java
```diff
@@ -0,0 +1,16 @@
...
+
+	public static boolean isScopePresentInWhiteList(String expression, List<Pattern> regexPatterns) {
```
dilipdevaraj-sfdc: is the intention of the variable regexPatterns or scopeRegexPatterns ?

ArgusCore/src/main/java/com/salesforce/dva/argus/util/AlertUtils.java
sundeepsf: yes, that is possible in theory, but in practice should be very rare. The way I am doing the match by putting : before and after the regex should minimize a lot of the such cases. Also, the only side effect, even if there is a wrong match is that the alert stays enabled, which by itself is not a really that big of an issue. If user complains, we can always fine tune the regex accordingly, depending on the situation.

PR #817, opened 2018-07-17
Long support for metric data value and trigger threshold

adelaidechambers:
Allow users to input metric data with long data values to Argus and long alert trigger threshold values. This way large values can be maintained precisely and triggers can be appropriately fired based off of these larger values.


nkunal:
@kgowdru @adelaidechambers  This is a fairly large change. We can continue with the review of this PR and incorporate feedback, but this should not get merged (and released) in our next release. Lets use this PR as an opportunity to come up with a set of integration/end-to-end/regression testing framework for Argus in general

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/NumberOperations.java
```diff
@@ -0,0 +1,862 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.salesforce.dva.argus.system.SystemAssert;
+
+public class NumberOperations {
```
adelaidechambers: Important class: I centralized all arithmetic operations on Numbers here to avoid repeating code

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/NumberOperations.java
```diff
@@ -0,0 +1,862 @@
...
+		}
+		return doubles;
+	}
+	
+	/**
+	 * @param map a map from long keys to number values
+	 * @return a map from long keys to double values, where the value is the double contained in map
+	 * @throws IllegalArgumentException if one of the values in map is not a double. If one of the values in map is null, will preserve it as null.
+	 */
+	public static Map<Long, Double> getMapAsDoubles(Map<Long, Number> map) throws IllegalArgumentException {
```
adelaidechambers: This method is useful for the transforms that are only supported when all datapoint values are doubles.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Trigger.java
```diff
@@ -88,22 +90,45 @@
...
 		public void serialize(Trigger trigger, JsonGenerator jgen, SerializerProvider provider) throws IOException, JsonProcessingException {
-			
 			jgen.writeStartObject();
 			
 			jgen.writeStringField("id", trigger.getId().toString());
 			jgen.writeStringField("name", trigger.getName());
 			jgen.writeStringField("type", trigger.getType().name());
-			jgen.writeNumberField("threshold", trigger.getThreshold().doubleValue());
+			
+			ValueType type = ValueType.value(trigger.getThreshold());
```
adelaidechambers: Important serialization code

dilipdevaraj-sfdc:
@adelaidechambers 
pls @mention reviewers so they can provide any review comments.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Trigger.java
```diff
@@ -88,22 +90,45 @@
...
+			switch (type) {
+				case INT:
```
dilipdevaraj-sfdc: what is the usecase for having INT when we already have LONG and DOUBLE.
If so is there any need for FLOAT ?

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/NumberOperations.java
dilipdevaraj-sfdc: It would be good to have relevant java doc for all public classes and public methods

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/NumberOperations.java
```diff
@@ -0,0 +1,862 @@
...
+	/**
+	 * Computes the approximate data value at a given percentile. Uses linear
+	 * interpolation between data values to approximate the value. Same algorithm as used in
+	 * org.apache.commons.math3.stat.descriptive.rank.Percentile.
+	 * 
+	 * @param vals the data values to analyze. Must be nonempty, cannot contain any null values.
+	 * @param p the percentile to use. Must be in (0, 100].
+	 * @return the approximate value at the pth percentile of vals.
+	 
+	public static Number percentile(Number[] vals, Number p) {
```
dilipdevaraj-sfdc: is this code meant to be there  since it is commented out?

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/NumberOperations.java
```diff
@@ -0,0 +1,862 @@
...
+	/*
+	/**
+	 * Computes the sample variance from a set of data.
+	 * 
+	 * @param vals the data values
+	 * @return the sample variance of vals
+	 
+	public static Number variance(Number[] vals) {
+		if (vals.length == 1) {
+			return 0;
```
dilipdevaraj-sfdc: same here commented out code ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/DefaultMetricService.java
```diff
@@ -126,8 +126,8 @@ protected DefaultMetricService(MonitorService monitorService, Provider<MetricRea
 				metrics.addAll(reader.parse(expression, relativeTo, Metric.class));
 				expandedTimeSeriesRange = reader.getExpandedTimeSeriesRange();
 				queryTimeWindow = reader.getQueryTimeWindow();
-				numDiscoveryResults += reader.getNumDiscoveryResults();
```
dilipdevaraj-sfdc: This change should be removed it should get the populated results back from reader object

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/AnomalyDetectionGaussianDensityTransform.java
```diff
@@ -80,5 +80,20 @@ public double calculateAnomalyScore(double value) {
         return -1.0 * Math.log(probabilityDensity);
     }
 
+    /*
+    @Override
```
dilipdevaraj-sfdc: If you feel there is unused commented code please delete it. 
For unsupported cases like Long type in Anomaly transforms - if  is not easily supported we can document it later.

ArgusWeb/app/js/directives/charts/lineChart.js
```diff
@@ -119,7 +119,7 @@ angular.module('argus.directives.charts.lineChart', [])
 							return $scope.menuOption;
 						}
 					},
-					templateUrl: 'js/templates/modals/chartOptions.html',
+					template: require('../../templates/modals/chartOptions.html'),
```
dilipdevaraj-sfdc: not sure why UI javascript changes are showing up. Can you please resync your PR branch to latest develop, so only relevant file diff is shown.

ArgusSDK/src/test/java/com/salesforce/dva/argus/sdk/MetricServiceTest.java
```diff
@@ -74,9 +74,10 @@ private Metric _constructMetric() throws JsonProcessingException {
         Metric result = new Metric();
         Map<String, String> fields = new TreeMap<>();
         Map<String, String> tags = new TreeMap<>();
-        Map<Long, Double> datapoints = new TreeMap<>();
+        Map<Long, Number> datapoints = new TreeMap<>();
```
dilipdevaraj-sfdc: Do we have a test which does something like
Case 1
// datapoint as long
putDatapoints(timestamp, 101)
getDatapoints
// assert value is exact precision as long 101 and not 101.0

similarly
Case2
// datapoint as double
putDatapoints(timestamp, 101.1)
getDatapoints
// assert value is exact precision as double 101.1 and not 101

Case3
Long number with greater than 53 bits (so precision is not last as with old implementation), and returned precision should be exact.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Trigger.java
```diff
@@ -132,11 +157,22 @@ public Trigger deserialize(JsonParser jp, DeserializationContext ctxt) throws IO
 			TriggerType type = TriggerType.fromString(rootNode.get("type").asText());
 			trigger.setType(type);
 			
-			Double threshold = rootNode.get("threshold").asDouble();
+			BigDecimal threshold;
+			if (rootNode.get("thresholdLong") != null) {
```
dilipdevaraj-sfdc: just a generic comment on use of BigDecimal - it will improve precision versus double when doing mathematical operations but at the same time it will lead to lower performance for certain large number operations.
so when we are chaining multiple transforms we need to ensure performance is not drastically affected due to this.

dilipdevaraj-sfdc:
@adelaidechambers 
Just for your reference
https://github.com/salesforce/Argus/commit/7ebb84890f22e50d71c5cb3f2fece80e22e11f12#diff-2b217964838665f96a303eb8bf467e07

previously value was a String but this would occupy more byte space and numeric calculations may not be precise all the time. Hence it was converted to double so it occupies a fixed 8 byte space.

dilipdevaraj-sfdc:
@nkunal  @colbyguan  @sundeepsf  @rajsarkapally  @asinghsfdc 
Can you also pls review this PR, so @adelaidechambers  can have sufficient time to incorporate feedback and do necessary testing.  
We can then make a call of how and when we want to deploy this feature change.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/NumberOperations.java
adelaidechambers: The commented out code would be used for implementing some of the transforms that are not currently supported with long values (it's part of future work). I'd written this code to replace some external API calls that we can't use with longs. We decided not to include those transforms here, since that would require significantly more testing for correctness. I left it in since I'd already written it and can be used later, but if it's better I can remove it.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Trigger.java
```diff
@@ -88,22 +90,45 @@
...
 			jgen.writeStringField("id", trigger.getId().toString());
 			jgen.writeStringField("name", trigger.getName());
 			jgen.writeStringField("type", trigger.getType().name());
-			jgen.writeNumberField("threshold", trigger.getThreshold().doubleValue());
+			
+			ValueType type = ValueType.value(trigger.getThreshold());
+			switch (type) {
+				case INT:
+				case LONG:
+					jgen.writeNumberField("thresholdLong", trigger.getThreshold().longValue());
```
adelaidechambers: Important serialization/deserialization code

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Trigger.java
adelaidechambers: I'm only using BigDecimal for the trigger thresholds because of JPA rounds Number like double, but maintains the precision for BigDecimal. I thought that the trigger threshold value was only used to compare to some other value, which would be calculated through an arbitrary number of transforms on metric data, containing Number values. How would the trigger threshold be used in chaining the transforms?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/DiscoveryService.java
```diff
@@ -49,7 +49,7 @@
 	
 	/** This should be a configuration. For now, this is how we reached on a value of 2M. 
 	 *  A datapoint in Argus is a tuple containing a Long timestamp (8 bytes with some additional Java Wrapper Class bytes) 
-	 *  and a Double value (8 bytes with some additional Java Wrapper Class bytes). We would then consider a datapoint to 
+	 *  and a Number value (8 bytes with some additional Java Wrapper Class bytes). We would then consider a datapoint to
```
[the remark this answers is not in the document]
adelaidechambers: I think this is still true. From what I've found, the data value will still just be the size of a long/double, which is 8 bytes, plus the class wrapper.

ArgusSDK/src/test/java/com/salesforce/dva/argus/sdk/MetricServiceTest.java
adelaidechambers: I've run all of these tests manually, but I can add them to this as well.

ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/dto/EntityDTO.java
```diff
@@ -71,9 +75,14 @@
      */
     public static <D extends EntityDTO, E extends JPAEntity> D createDtoObject(Class<D> clazz, E entity) {
         D result = null;
+        
+        BigDecimal defaultValue = null;
```
adelaidechambers: This works around a known bug for BigDecimal where BeanUtils throws an exception for any null values.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Trigger.java
adelaidechambers: I'd added the INT option to ValueType because sometimes we want to compare things like the size of some list, etc. to another value, so it compares the int to a Number within NumberOperations. It's true that shouldn't come up within Trigger, so I can remove that from this class

ArgusSDK/src/test/java/com/salesforce/dva/argus/sdk/MetricServiceTest.java
dilipdevaraj-sfdc: Its always a good practice to write unit tests if possible in addition to just manual tests. That way when there is a future release a new feature does not cause regression, since this will be captured via unit test.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Trigger.java
dilipdevaraj-sfdc: ok I see. Trigger threshold is only used for comparison when evaluating the computed metric expression and would not be involved in transforms so we should be good.

Maybe you can add comment in java code "BigDecimal used here since JPA rounds Number like double, but maintains the precision for BigDecimal." , since this is not so obvious.

ArgusWeb/app/js/argusConfig.js
```diff
@@ -29,74 +29,74 @@ angular.module('argus.config', [])
 		paginationTemplateProvider.setPath('node_modules/angular-utils-pagination/dirPagination.tpl.html');
 		$routeProvider.
 			when('/beta', {
-				templateUrl: 'js/templates/beta.html',
+				template: require('./templates/beta.html'),
```
colbyguan: I recall these diffs are from some wonky merge-from-master; we could look into doing a `git reset` of these files to this branch's base commit to keep the PR clean

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/NumberOperations.java
```diff
@@ -0,0 +1,803 @@
...
+		return vals.stream()
+				.reduce(0, (a, b) -> add(a, b));
+	}
+	
+	/**
+	 * @param vals the list of numbers
+	 * @return the list of doubles contained in vals. If one element of vals is null, will preserve it as null.
+	 * @throws IllegalArgumentException if one of the numbers in vals is not a double
+	 */
+	public static List<Double> getListAsDoubles(List<Number> vals) throws IllegalArgumentException {
```
colbyguan: `IllegalArgumentException` is unchecked, so I believe there's no need to indicate it in the method signature; javadoc is good enough. 
Also, from a quick Command+F it seems this method is always used inside a `try-catch` block. It seems cleaner to just let the exception be thrown at this level with a message like `Expected class type of List elements to be Doubles` and have it be called without try-catch; the stack trace will say what transform it failed on.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/PercentileValueReducerOrMapping.java
```diff
@@ -109,15 +123,15 @@ public String name() {
     private Map<Long, Double> _calculateNthPercentileForOneMetric(Map<Long, Double> originalDatapoints, Double percentileValue) {
     	
 	    Map<Long, Double> result = new TreeMap<>();
-	    for(Long timestamp : originalDatapoints.keySet()) {
+	    for(Long timestamp : new TreeMap<>(originalDatapoints).keySet()) { // want to get the earliest timestamp
```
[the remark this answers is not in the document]
adelaidechambers: The expected behavior of this transform should be to return a data entry with the timestamp of the  earliest timestamp  in the queried data and the pth percentile value. Using tree map ensures that the earliest timestamp is the one used -- otherwise it might just iterate through according to the hashed ordering, not the natural ordering. OriginalDatapoints is treated as a generic map here.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/ShiftValueMapping.java
```diff
@@ -48,22 +48,26 @@
...
 
-        for (Entry<Long, Double> entry : originalDatapoints.entrySet()) {
+        for (Entry<Long, Number> entry : originalDatapoints.entrySet()) {
+        	if (offset >= 0) {
+        		SystemAssert.requireArgument((entry.getKey() <= Long.MAX_VALUE - offset && entry.getKey() >= Long.MIN_VALUE), 
+        				"You are not allowed to shift like this, be nice to me!");
+        	} else {
+        		SystemAssert.requireArgument((entry.getKey() <= Long.MAX_VALUE && entry.getKey() >= Long.MIN_VALUE - offset),
+        				"You are not allowed to shift like this, be nice to me!");
+        	}
```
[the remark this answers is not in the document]
adelaidechambers: The old behavior was a NOP. The statement 'if offset positive or 0 then ensure key+offset <= Long.MAX' will always be true here. Key and offset are both treated as longs and summed into the long newTimestamp, so is guaranteed to be a long regardless of whether that long value has overflowed. Assuming offset is a valid long, though, and calculating key <= Long.MAX - offset could be violated.
For example, consider the following: let key = Long.MAX_VALUE, offset = 1. Then, key + offset = (Long.MAX_VALUE + 1) = Long.MIN_VALUE <= Long.MAX_VALUE. But, in the new check, key = Long.MAX_VALUE > Long.MAX_VALUE - 1 = Long.MAX_VALUE - offset. So,, this wouldn't cause an exception in the old loop but would in the new one. I assume those kinds of shifts were the purpose of the exception, so now it will catch them.
I added the check that the constants were within the bounds for a long value first since, in theory, they could be double values outside of the scope of a long value. I agree that these exceptions would be caused by user error, but thought that was largely the intention of the original exception anyways (similar to providing an illegal constant in the other transforms).

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/AnomalyDetectionKMeansTransform.java
```diff
@@ -171,7 +178,7 @@ private Metric predictAnomalies(Map<Long, Double> metricData) {
             }
         }
 
-        predictions.setDatapoints(predictionDatapoints);
+        predictions.setDatapoints(new HashMap<>(predictionDatapoints));
```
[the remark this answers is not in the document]
adelaidechambers: AnomalyDetectionKMeansTransform is one of transforms that is only supported for double values. I had predictionDatapoints as a map from long to double, so I had to cast it to a map of long to Number to resolve compilation issues. I went back through and updated predictionDatapoints to be a map from long to number here and in all of the other anomaly transforms.

dilipdevaraj-sfdc:
@adelaidechambers Can you please sign the CLA for the open source project.

PR #819, opened 2018-07-17
Change scopenames list to set to reduce duplicate record count at sta…

dilipdevaraj-sfdc:
…rtup time


dilipdevaraj-sfdc:
@nkunal pls review

PR #825, opened 2018-07-19
Customized argus alert trigger name

studanshu:
Integrating some of the common metric information into alert trigger name, like scope or device, so we can define our trigger name like this watchdog-trigger-${scope} or watchdog-trigger-${scope}-${device}.


ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/notifier/DefaultNotifier.java
```diff
@@ -95,11 +98,31 @@ public void sendNotification(NotificationContext notificationContext) {
...
     }
 
+    /*
+    * Finds all the templates like ${scope}, ${metric} and replaces it with the required fields.
+    * If no matches are found, nothing is done.
+    * */
+    protected void updateTriggerName(NotificationContext context) {
+        String newTriggerName = context.getTrigger().getName();
+        Metric triggeredMetric = context.getTriggeredMetric();
+        newTriggerName = newTriggerName.replaceAll("\\$\\{scope\\}", triggeredMetric.getScope());
```
sundeepsf: is it possible to do case insensitive match for the variables?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/notifier/DefaultNotifier.java
```diff
@@ -95,11 +98,31 @@ public void sendNotification(NotificationContext notificationContext) {
...
+        newTriggerName = newTriggerName.replaceAll("\\$\\{metric\\}", triggeredMetric.getMetric());
+        Map<String, String> tags = triggeredMetric.getTags();
+        Matcher m = Pattern.compile("\\$\\{.*?\\}").matcher(newTriggerName);
+        while (m.find()) {
+            String currentRegex = m.group(), currentTagKey = currentRegex.substring(2, currentRegex.length()-1);
+            if (tags.containsKey(currentTagKey))
+                newTriggerName = newTriggerName.replace(currentRegex, tags.get(currentTagKey));
```
sundeepsf: will this cause any problem if the same tagname is specified multiple times? Maybe not, but good to have a test case for that

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/notifier/DefaultNotifier.java
studanshu: @sundeepsf done.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/notifier/DefaultNotifier.java
studanshu: @sundeepsf added.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/notifier/DefaultNotifier.java
```diff
@@ -95,11 +98,31 @@ public void sendNotification(NotificationContext notificationContext) {
...
+    * */
+    protected void updateTriggerName(NotificationContext context) {
+        String newTriggerName = context.getTrigger().getName();
+        Metric triggeredMetric = context.getTriggeredMetric();
+        newTriggerName = newTriggerName.replaceAll("(?i)\\$\\{scope\\}", triggeredMetric.getScope());
+        newTriggerName = newTriggerName.replaceAll("(?i)\\$\\{metric\\}", triggeredMetric.getMetric());
+        Map<String, String> tags = triggeredMetric.getTags();
+        Matcher m = Pattern.compile("(?i)\\$\\{.*?\\}").matcher(newTriggerName);
+        while (m.find()) {
+            String currentRegex = m.group(), currentTagKey = currentRegex.substring(2, currentRegex.length()-1);
```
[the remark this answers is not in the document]
studanshu: @sundeepsf It's a good idea, I had thought about it. The only difficulty I find is in the number of queries to the Map of {tagKey,tagValue}. There will exists exponential possibility as the Map is case-sensitive.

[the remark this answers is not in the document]
studanshu: @sundeepsf Are the tag keys saved in database all in lowercase? I thought it's not the case. For eg, "tagkey" can be entered by user as {"TagKey","tAgkEy","TAGKEY"}, etc.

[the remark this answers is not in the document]
sundeepsf: No, they are not stored that way. I meant construct another map with lowercase keys

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/notifier/DefaultNotifier.java
```diff
@@ -95,11 +98,40 @@ public void sendNotification(NotificationContext notificationContext) {
...
+        updateTriggerName(notificationContext);
         _createAnnotation(notificationContext, additionalFields);
         sendAdditionalNotification(notificationContext);
         _dispose();
     }
 
+    private  Map<String, String> getLowerCaseTagMap(final Map<String, String> tags) {
+        Map<String, String> lowerCaseTagMap = new HashMap<>();
+        for (String originalTags: tags.keySet()) {
+            lowerCaseTagMap.put(originalTags.toLowerCase(), originalTags);
```
sundeepsf: should you not be putting the tag value into this map?

studanshu:
Yes. It would be much more optimized. Did.
Regards
Sudhanshu Bahety
Associate Member of Technical Staff, Salesforce com

PR #826, opened 2018-07-19
Add new scope+metric index

naveenreddykarri:
PR is read to sign off.
Added unit tests.
Tested out the changes in dev cluster

### DIFF
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeAndMetricOnlySchemaRecord.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeAndMetricOnlySchemaRecord.java
@@ -100,7 +100,7 @@ public class ScopeAndMetricOnlySchemaRecord {
         return MessageFormat.format("ScopeAndMetricOnlySchemaRecord (Scope = {0} Metric = {1}", scope, metric);
     }
 
-    public static String print(ScopeAndMetricOnlySchemaRecord msr) {
+    public static String getFieldsAsString(ScopeAndMetricOnlySchemaRecord msr) {
 
         StringBuilder sb = new StringBuilder(msr.getScope());
         sb.append(":");
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeAndMetricOnlySchemaRecord.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeAndMetricOnlySchemaRecord.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2016, Salesforce.com, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of Salesforce.com nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.salesforce.dva.argus.entity;
+
+import com.google.common.base.Objects;
+
+import java.text.MessageFormat;
+
+/**
+ * Represents a search result row for scope and metric name only discovery queries.
+ *
+ * @author  Naveen Reddy Karri (nkarri@salesforce.com)
+ */
+public class ScopeAndMetricOnlySchemaRecord {
+
+    private String scope;
+    private String metric;
+
+    public ScopeAndMetricOnlySchemaRecord(String scope, String metric) {
+
+        setScope(scope);
+        setMetric(metric);
+    }
+
+    public String getScope() {
+        return scope;
+    }
+
+    public void setScope(String scope) {
+        this.scope = scope;
+    }
+
+    public String getMetric() {
+        return metric;
+    }
+
+    public void setMetric(String metric) {
+        this.metric = metric;
+    }
+
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+
+        result = prime * result + ((scope == null) ? 0 : scope.hashCode());
+        result = prime * result + ((metric == null) ? 0 : metric.hashCode());
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+
+        ScopeAndMetricOnlySchemaRecord other = (ScopeAndMetricOnlySchemaRecord) obj;
+
+        return Objects.equal(scope, other.scope) && Objects.equal(metric, other.metric);
+    }
+
+    @Override
+    public String toString() {
+        return MessageFormat.format("ScopeAndMetricOnlySchemaRecord (Scope = {0} Metric = {1}", scope, metric);
+    }
+
+    public static String print(ScopeAndMetricOnlySchemaRecord msr) {
+
+        StringBuilder sb = new StringBuilder(msr.getScope());
+        sb.append(":");
+        sb.append(msr.getMetric());
+
+        return sb.toString();
+    }
+}
+/* Copyright (c) 2016, Salesforce.com, Inc.  All rights reserved. */
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
@@ -14,6 +14,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -39,18 +40,22 @@ public abstract class AbstractSchemaService extends DefaultService implements Sc
 	private static final int DAY_IN_SECONDS = 24 * 60 * 60;
 	private static final int HOUR_IN_SECONDS = 60 * 60;
 
-	/* Have two separate bloom filters one for metrics schema and another for scope names schema.
-	 * Since scopes will continue to repeat more often on subsequent kafka batch reads, we can easily check this from the  bloom filter for scopes only. 
+	/* Have three separate bloom filters one for metrics schema, one only for scope names schema and one only for scope name and metric name schema.
+	 * Since scopes will continue to repeat more often on subsequent kafka batch reads, we can easily check this from the  bloom filter for scopes only.
 	 * Hence we can avoid the extra call to populate scopenames index on ES in subsequent Kafka reads.
+	 * The same logic applies to scope name and metric name schema.
 	 */
 	protected static BloomFilter<CharSequence> bloomFilter;
 	protected static BloomFilter<CharSequence> bloomFilterScopeOnly;
+	protected static BloomFilter<CharSequence> bloomFilterScopeAndMetricOnly;
 	private Random rand = new Random();
 	private int randomNumber = rand.nextInt();
 	private int bloomFilterExpectedNumberInsertions;
 	private double bloomFilterErrorRate;
 	private int bloomFilterScopeOnlyExpectedNumberInsertions;
-	private double bloomFilterScopeOnlyErrorRate;	
+	private double bloomFilterScopeOnlyErrorRate;
+	private int bloomFilterScopeAndMetricOnlyExpectedNumberInsertions;
+	private double bloomFilterScopeAndMetricOnlyErrorRate;
 	private final Logger _logger = LoggerFactory.getLogger(getClass());
 	private final Thread _bloomFilterMonitorThread;
 	protected final boolean _syncPut;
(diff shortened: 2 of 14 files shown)


ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -124,24 +141,37 @@ public void put(List<Metric> metrics) {
...
+			key = constructScopeAndMetricOnlyKey(scopeName, metricName);
+			found = bloomFilterScopeAndMetricOnly.mightContain(key);
+			if(!found) {
+				scopesAndMetricsNamesToPut.add(new Metric(scopeName, metricName));
+			}
 		}
 
-		implementationSpecificPut(metricsToPut, scopesToPut);
+		List<Metric> scopesAndMetricsToPut = new ArrayList<>(scopesAndMetricsNamesToPut);
+		implementationSpecificPut(metricsToPut, scopesToPut, scopesAndMetricsToPut);
```
nkunal: you should use a Set for scopesAndMetricsNamesToPut, similar to scopesToPut.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -124,24 +141,37 @@ public void put(List<Metric> metrics) {
...
 	/*
 	 * Calls the implementation specific write for indexing the records
 	 *
 	 * @param  metrics    The metrics metadata that will be written to a separate index.
 	 * @param  scopeNames The scope names that that will be written to a separate index.
+	 * @param  scopesAndMetricNames The scope and metric names that that will be written to a separate index.
 	 */
-	protected abstract void implementationSpecificPut(List<Metric> metrics, Set<String> scopeNames);
+	protected abstract void implementationSpecificPut(List<Metric> metrics, Set<String> scopeNames,
+													  List<Metric> scopesAndMetricNames);
```
nkunal: use a Set for scopesAndMetricNames

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -195,18 +225,22 @@ protected String constructKey(Metric metric, Entry<String, String> tagEntry) {
...
 
-		if(tagk != null) {
+		if(!StringUtils.isEmpty(tagk)) {
 			sb.append('\0').append(tagk);
 		}
 
-		if(tagv != null) {
+		if(!StringUtils.isEmpty(tagv)) {
 			sb.append('\0').append(tagv);
 		}
```
nkunal: The sb.append('\0').append(randomNumber); needs to appended to each Key so that we get uniqueness across multiple schema client and we work around the bloom filter false positive error rate.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -261,11 +295,26 @@ protected int getNumHoursUntilTargetHour(int targetHour){
...
+
+		BLOOMFILTER_SCOPE_ONLY_EXPECTED_NUMBER_INSERTIONS("service.property.schema.bloomfilter.scope.only.expected.number.insertions", "1000000"),
 		BLOOMFILTER_SCOPE_ONLY_ERROR_RATE("service.property.schema.bloomfilter.scope.only.error.rate", "0.00001"),
+
+		/*
+		 * Estimated Filter Size using https://hur.st/bloomfilter/?n=10000000&p=1.0E-5&m=&k=
+		 * 28.56MiB
+		*/
+
+		BLOOMFILTER_SCOPE_AND_METRIC_ONLY_EXPECTED_NUMBER_INSERTIONS("service.property.schema.bloomfilter.scope.and.metric.only.expected.number.insertions", "10000000"),
```
nkunal: in production you should run with 500 Million.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -16,7 +16,10 @@
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.function.Supplier;
 
+import com.google.common.hash.BloomFilter;
+import com.salesforce.dva.argus.entity.*;
```
nkunal: Use specific imports

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -353,7 +398,8 @@ protected void implementationSpecificPut(List<Metric> metrics, Set<String> scope
 					requestBody.put("scroll_id", scrollID);
 					requestBody.put("scroll", KEEP_SCROLL_CONTEXT_OPEN_FOR);
 
-					response = _esRestClient.performRequest(HttpMethod.POST.getName(), requestUrl, Collections.emptyMap(), new StringEntity(new ObjectMapper().writeValueAsString(requestBody)));
+					response = _esRestClient.performRequest(HttpMethod.POST.getName(), requestUrl, Collections.emptyMap(),
+							new StringEntity(new ObjectMapper().writeValueAsString(requestBody)));
```
nkunal: is there an extra newline here ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -621,11 +667,11 @@ protected void implementationSpecificPut(List<Metric> metrics, Set<String> scope
 		}
 	}
 
-	private void _upsert(List<MetricSchemaRecord> records) {
+	private void _upsert(List<MetricSchemaRecord> records, String indexName, String indexType, BloomFilter<CharSequence> bloomFilter) {
```
nkunal: indexType is actually called docType (in elasticsearch terminology )

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -1188,10 +1232,21 @@ public void setName(String name) {
...
+
 		/** Name of scope only index */
 		ELASTICSEARCH_SCOPE_INDEX_NAME("service.property.schema.elasticsearch.scope.index.name", "scopenames"),
 		/** Type within scope only index */
-		ELASTICSEARCH_SCOPE_TYPE_NAME("service.property.schema.elasticsearch.scope.type.name", "scope_type");
+		ELASTICSEARCH_SCOPE_TYPE_NAME("service.property.schema.elasticsearch.scope.type.name", "scope_type"),
+
+		/** Replication factor for scope and metric names */
+		ELASTICSEARCH_NUM_REPLICAS_FOR_SCOPE_AND_METRIC_INDEX("service.property.schema.elasticsearch.num.replicas.for.scope.metric.index", "1"),
+		/** Shard count for scope and metric names */
```
nkunal: should we name the props as elasticsearch.num.replicas.for.scopeandmetric.index ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -124,24 +141,37 @@ public void put(List<Metric> metrics) {
...
-				scopesToPut.add(metric.getScope());
-			}			
+				scopesToPut.add(scopeName);
+			}
+
+			// Check scope and metric only bloom filter
+			key = constructScopeAndMetricOnlyKey(scopeName, metricName);
+			found = bloomFilterScopeAndMetricOnly.mightContain(key);
+			if(!found) {
+				scopesAndMetricsNamesToPut.add(new Metric(scopeName, metricName));
```
nkunal: We should probably create a new class ScopeAndMetricOnlySchemaRecord (similar to ScopeOnlySchemaRecord) and have only scope.metric field names. There will be corresponding JsonSerielizer for it that produce only those two fields, with md5 hash id of scope+metric which will be used as _id field for ES docs. thoughts ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
naveenreddykarri: yes added it since it becomes difficult to read if > 120 chars when doing github reviews.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
naveenreddykarri: scopesAndMetricsNamesToPut is a Set to remove duplictes, I finally convert it to scopesAndMetricsToPut which is a list, so that I can reuse the rest of the functions which act on List<Metric>

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
naveenreddykarri: scopesAndMetricsNamesToPut is a Set to remove duplictes, I finally convert it to scopesAndMetricsToPut which is a list, so that I can reuse the rest of the functions which act on List

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
naveenreddykarri: It will be appended to each key on line 249. sb.append('\0').append(randomNumber);

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
naveenreddykarri: I had 10 million. # of metrics + scopenames, what is the today's value. We should probably have 10 times today's value? 500 million, the bloom filter will be very huge. 1.39 GB.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
naveenreddykarri: fixed

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
naveenreddykarri: changed scope.metric to scopeandmetric. Should I also remove service.property.schema.?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
naveenreddykarri: I initially had ScopeAndMetricOnlySchemaRecord. Then I saw that we end up duplicating a lot of code, just because they work on different objects, call different functions(different functions since again they work on different objects) etc. One option is we could use generics to simplify all that. I also noticed that MetricSchemaRecord was designed to handle the case when only scope and metric name are present in terms of serialization, deserialization etc. I haven't completed tested it yet, but if I missed something and we need a additional class, I will add it. The Json Serializer should generate only those two fields when other fields are null. I will add tests to verify that.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
nkunal: For metrics+scope our volume is 132 Million. I think we should have the scopemetric BF with at least 200 MM. this 132 MM are distributed ununiformaly acroos 10 schema clients
We run with 64 GB per instance, so there is plenty of mem available. Out of that 10 Gb is used by the metadata_index BF. You have another 30gb available for BF if push comes to shove

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -124,24 +142,36 @@ public void put(List<Metric> metrics) {
...
-		implementationSpecificPut(metricsToPut, scopesToPut);
+		implementationSpecificPut(metricsToPut, scopesToPut, scopesAndMetricsNamesToPut);
 	}
 
 	/*
 	 * Calls the implementation specific write for indexing the records
 	 *
 	 * @param  metrics    The metrics metadata that will be written to a separate index.
 	 * @param  scopeNames The scope names that that will be written to a separate index.
+	 * @param  scopesAndMetricNames The scope and metric names that that will be written to a separate index.
```
dilipdevaraj-sfdc: remove extra that

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -124,24 +142,36 @@ public void put(List<Metric> metrics) {
...
 
-		implementationSpecificPut(metricsToPut, scopesToPut);
+		implementationSpecificPut(metricsToPut, scopesToPut, scopesAndMetricsNamesToPut);
 	}
 
 	/*
 	 * Calls the implementation specific write for indexing the records
 	 *
 	 * @param  metrics    The metrics metadata that will be written to a separate index.
 	 * @param  scopeNames The scope names that that will be written to a separate index.
```
dilipdevaraj-sfdc: same here

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -16,7 +16,11 @@
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.function.Supplier;
 
+import com.google.common.hash.BloomFilter;
+import com.salesforce.dva.argus.entity.*;
```
dilipdevaraj-sfdc: Can specific import be used here

ArgusCore/src/test/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaServiceTest.java
```diff
@@ -107,10 +258,89 @@ public void testGetUniqueUsingMetricSchemaIndex() throws IOException {
...
+        assertFalse(queryForMetric.isQueryOnlyOnScope());
+        assertFalse(queryForMetric.isQueryOnlyOnScopeAndMetric());
+    }
+
+    private String convertToPrettyJson(String jsonString) {
+        JsonParser parser = new JsonParser();
+        Gson gson = new GsonBuilder().setPrettyPrinting().create();
+
+        JsonElement el = parser.parse(jsonString);
+        return gson.toJson(el); // done
```
dilipdevaraj-sfdc: remove // done

ArgusCore/src/test/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaServiceTest.java
```diff
@@ -107,10 +258,89 @@ public void testGetUniqueUsingMetricSchemaIndex() throws IOException {
...
+
+        assertEquals(metricQuery1, queryJson);
         assertEquals("/metadata_index/metadata_type/_search", requestUrl);
 
         assertFalse(queryForMetric.isQueryOnlyOnScope());
+        assertFalse(queryForMetric.isQueryOnlyOnScopeAndMetric());
+    }
+
+    @Test
+    public void testGetUniqueUsingMetricSchemaIndex2() throws IOException {
```
dilipdevaraj-sfdc: Its not very clear what we are testing just from the method names
testGetUniqueUsingMetricSchemaIndex1
testGetUniqueUsingMetricSchemaIndex2

we are using two different metric queries but are still searching same index in bothe tests /metadata_index/metadata_type/_search

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/SchemaRecordList.java
```diff
@@ -0,0 +1,42 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.JsonNodeType;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+public class SchemaRecordList {
```
dilipdevaraj-sfdc: Can you add javadoc
Did you run mvn javadoc:javadoc successfully ?

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeAndMetricOnlySchemaRecord.java
```diff
@@ -0,0 +1,112 @@
...
+
+        return Objects.equal(scope, other.scope) && Objects.equal(metric, other.metric);
+    }
+
+    @Override
+    public String toString() {
+        return MessageFormat.format("ScopeAndMetricOnlySchemaRecord (Scope = {0} Metric = {1}", scope, metric);
+    }
+
+    public static String print(ScopeAndMetricOnlySchemaRecord msr) {
```
dilipdevaraj-sfdc: Please rename method from print to maybe getFieldsAsString
note this is a static method which is different from the regular toSting() instance method.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/SchemaRecordList.java
naveenreddykarri: yes. There were other errors in classes that I have not touched, I did not fix them. But none in these classes.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeAndMetricOnlySchemaRecord.java
naveenreddykarri: fixed

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
nkunal: usually in such situations, each param to the function goes on its own line. that way it is easier to track what params are passed to each function and if there "new" or another function call as param to the orig function, then just do the new or function call prior and pass the var to the original function

PR #835, opened 2018-07-23
Alerts pagination (Part 1, new API in ArgusCore, ArgusWS)

dongpujin:
Add endpoints below to facilitate alert pagination. This PR only introduce new APIs will be used by pagination, but does not have any impact on existing logic at this moment. 
- /alerts/meta/user: return a page of user's alerts; 
- /alerts/meta/user/count: returns total count of user's alerts;
- /alerts/meta/shared: return a page of all shared alerts; 
- /alerts/meta/shared/count: returns total count of all shared alerts; 
- /alerts/meta/privileged: returns all non-shared alerts if privileged user; 
- /alerts/meta/privileged/count: returns total count of all non-shared alerts if privileged user; 

Unit tests added. 
Tested locally by hooking up with some temporary UI glue code, and pagination works where a batch of alerts are fetched from DB when flip a page. 

TODO: 
- Create search queries. 
- UI changes to call new APIs.


ArgusCore/src/main/java/com/salesforce/dva/argus/service/AlertService.java
```diff
@@ -136,6 +136,24 @@
...
 	List<Alert> findAlertsByOwner(PrincipalUser owner, boolean metadataOnly);
+	
+	/**
+	 * Returns a list of alerts for an owner with given limit and offset.
+	 *
+	 * @param owner The owner to return alerts for. Cannot be null.
+	 *
+	 * @return The list of alerts.
+	 */
+	List<Alert> findAlertsByOwnerPaged(PrincipalUser owner, boolean metadataOnly, Integer limit, Integer offset);
```
sundeepsf: There is a a lot of duplicate logic between the new methods and the existing methods. Since limit and offset can be added as extra params to the jpa query using setMaxResults and setFirstResult methods respectively, can't we just add additional params to existing method and refactor the caller code slightly to pass in null when these params are not required?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/AlertService.java
```diff
@@ -136,6 +136,24 @@
...
+	
+	/**
+	 * Count the total number of alerts for the owner.
+	 *
+	 * @param owner The owner of the alerts. Cannot be null.
+	 *
+	 * @return The total number of owner's alert.
+	 */
+	int countAlertsByOwner(PrincipalUser owner);
```
sundeepsf: can we make this count API take in a list of params and construct the query dynamically based on the params? This way we would just need one API instead of three

ArgusCore/src/main/java/com/salesforce/dva/argus/service/AlertService.java
sundeepsf: Yes, thats fine. You can have separate method interfaces, but as long as the underlying implementation class is reusing the code should be good. Like I mentioned, you can directly set the limit and offset directly on the jpaquery before execution. From what I see , that probably would be the only change

dongpujin: I too wish it was as simple as adding limit/offset. However, the complication is currently a request from the UI runs multiple underlying jpa queries and assemble results into a set to dedup, like this method, https://github.com/salesforce/Argus/blob/develop/ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/resources/AlertResources.java#L112-L122. 

Simply adding limit and offset to each under jpa query will not make pagination work holistically. What I was doing was essentially introduce atomic APIs, where each APIs is a single paginated jpa query, responsible for each alert tab. 

I also consolidated the count API into one, and some duplicated code. Please review when you get a chance. Thanks :-)

sundeepsf: Hmm, want to understand the problem a bit more. Can we have a quick meeting to discuss this tomorrow? Its easier to talk and resolve things

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/AlertsCountContext.java
```diff
@@ -0,0 +1,111 @@
...
+	// Boolean flag of whether counting private (non-shared) alerts for
+	// privileged user
+	private boolean countPrivateAlerts;
+
+	// Current owner
+	private PrincipalUser owner;
+
+	// Constructor
+	private AlertsCountContext(AlertsCountContextBuilder builder) {
+		this.countUserAlerts = builder.userAlerts;
```
sundeepsf: are these options mutually exclusive? Like if we specify sharedAlert=true can we also specify privateAlerts=true? If that is the case maybe you need an input check, like for example if privateAlerts is true then owner should be present

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
```diff
@@ -827,6 +835,59 @@ public int alertCountByStatus(boolean enabled) {
...
+	public int countAlerts(AlertsCountContext context) {
+		requireNotDisposed();
+		
+		if (context == null) {
+			return 0;
+		}
+
+		// Count total number of shared alerts for the shared alerts tab
+		if (context.isCountSharedAlerts()) {
+			return Alert.countSharedAlerts(_emProvider.get());
```
sundeepsf: wondering if we can return total count based on all the params passed in? Like we can have a usecase where as a user I want total count of all alerts that I can see. This will include my private alerts as well as all the alerts shared with me. I know we don't use it like that in today's UI, but it might be a valid usecase

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Alert.java
```diff
@@ -149,6 +151,31 @@
 			@NamedQuery(
 					name = "Alert.getSharedAlertsByOwner", 
 					query = "SELECT a from Alert a where a.owner = :owner AND a.shared = true AND a.id not in (SELECT jpa.id from JPAEntity jpa where jpa.deleted = true)"
+					),
+			// Paginated queries
+			@NamedQuery(
```
sundeepsf: There is no difference between these paged queries and query like Alert.findByOwner. Can we get rid of these?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/AlertService.java
```diff
@@ -290,7 +307,24 @@
 	 * @param notifications  The notifications to update.	
 	 */
 	void updateNotificationsActiveStatusAndCooldown(List<Notification> notifications);
-
+	
+	/**
+	 * Find a list of private alerts (non-shared alerts) for the given privileged user.
+	 *
+	 * @param owner The owner to filter on.
```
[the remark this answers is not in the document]
dongpujin: Sure thing. Thanks.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
dongpujin: Current iteration of the count API is not that intelligent, and the count API's main purpose for now is the UI can call and figure out alert count in order to correctly populated page numbers. For above usecase, I think it will be easy to extend. If it's needed, we can update the method to check both private and user alert condition in the context, if  both are true, then it can run another count query to count distinct user and shared alerts.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Alert.java
dongpujin: You are right, we don't need named queries. Let me remove them to keep it simple.

PR #836, opened 2018-07-23
Support displaying empty time range at the beginning or the end of a graph

xizi-xu:
- Added an option in the chart option to turn on displaying of the empty time range of a graph, instead of only showing the time range with data
screen shot 2018-07-23 at 4 01 35 pm

- Add an additional alert condition to comply backend


ArgusWeb/app/js/controllers/alertsDetail.js
```diff
@@ -244,7 +244,8 @@ angular.module('argus.controllers.alerts.detail', ['ngResource'])
 			{label: '>=', value: 'GREATER_THAN_OR_EQ'},
 			{label: '<=', value: 'LESS_THAN_OR_EQ'},
 			{label: '<>', value: 'BETWEEN'},
-			{label: '><', value: 'NOT_BETWEEN'}
+			{label: '><', value: 'NOT_BETWEEN'},
+			{label: 'no data', value: 'NO_DATA'}
```
xizi-xu: @sundeepsf

xizi-xu: When user sets an alert, he enters the query expression in the UI and that's captured in the alert object as shown [here in the code](https://github.com/xizi-xu/Argus/blob/c784e556679e9d15a7ec9012f306e6b2bf89f2c7/ArgusWeb/app/js/templates/alert-detail.html#L30). The raw expression is sent as the part of the alert object via API to the server.

PR #838, opened 2018-07-24
Bug filltransform

studanshu:
When we execute this transform - FILL(#-115m#, #-15m#, #10m#, #0m#, #0#) , it is producing wrong start and end time intervals. There seems to be some normalization being done to align the intervals to a value divisible by 10. This is incorrect and the timewindow should exactly match the start and end times of the query


ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/FillTransform.java
```diff
@@ -149,11 +149,15 @@ private static long _parseTimeIntervalInSeconds(String interval) {
         SystemAssert.requireArgument(startTimestamp < endTimestamp, "End time must occure later than start time!");
         SystemAssert.requireArgument(windowSizeInSeconds >= 0, "Window size must be greater than ZERO!");
 
-        // snapping start and end time
-        long startSnapping = startTimestamp % (windowSizeInSeconds * 1000);
-        startTimestamp = startTimestamp - startSnapping;
-        long endSnapping = endTimestamp % (windowSizeInSeconds * 1000);
-        endTimestamp = endTimestamp - endSnapping;
+        boolean isDivsible = ((startTimestamp - endTimestamp) % (windowSizeInSeconds * 1000)) == 0;
```
sundeepsf: typo isDivisible

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/FillTransform.java
```diff
@@ -149,11 +149,15 @@ private static long _parseTimeIntervalInSeconds(String interval) {
...
+
+        // snapping start and end time if range is not a multiple of windowSize.
+        if (isDivsible == false) {
```
sundeepsf: you can also do if(!isDivisible)

ArgusCore/src/test/java/com/salesforce/dva/argus/service/metric/transform/FillTransformTest.java
```diff
@@ -478,8 +478,8 @@ public void testFillLineWithIntervalEqualsToFillRangeOffsetNegative() {
 
         Map<Long, Double> expected = new HashMap<Long, Double>();
 
-        expected.put(-1000L, 100.0);
-        expected.put(1000L, 100.0);
+        expected.put(0L, 100.0);
+        expected.put(2000L, 100.0);
```
sundeepsf: is there a test for indivisible step size?

studanshu: @sundeepsf yes.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/FillTransform.java
studanshu: Ok.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/FillTransform.java
studanshu: Changed.

dilipdevaraj-sfdc:
Post release please update the wiki to reflect the new fill functionality with an example
https://github.com/salesforce/Argus/wiki/Transforms#fill

PR #859, opened 2018-07-31
Rolling back the fill transform bug fix in 2.17 branch

sundeepsf:


dilipdevaraj-sfdc:
@sundeepsf 
Why are we rolling back the changes ?

sundeepsf:
@dilip-devaraj , there was a problem with the fix that it is not rounding the start and end values to the nearest minute.

dilipdevaraj-sfdc:
@sundeepsf 
If this is going to be deployed, as part of 2.17 release branch, just  for this PR can you increment all version files to 2.17.2.
(This is because maven central already has 2.17.1, and will not allow you to override artifacts).

Going forward only when the release branch is fully qualified we will release it to master so we don't need to keep incrementing version.

PR #847, opened 2018-07-27
Adding cts field

rajsarkapally:
@nkunal

PR #861, opened 2018-08-02
Add 'cts' field for scopemetricnames index | Add 'cts' field for scopenames index

naveenreddykarri:
Add 'cts' field for scopenames index
Add 'cts' field for scopemetricnames index
Fix bug with alertdefinition logging. 
Enabling using scopeandmetric.index.

### DIFF
--- a/ArgusClient/src/main/java/com/salesforce/dva/argus/client/SchemaCommitter.java
+++ b/ArgusClient/src/main/java/com/salesforce/dva/argus/client/SchemaCommitter.java
@@ -84,7 +84,7 @@ class SchemaCommitter extends AbstractCommitter {
                 Thread.currentThread().interrupt();
                 break;
             } catch (Throwable ex) {
-                LOGGER.info("Error occurred while committing metrics for schema records creation.", ex);
+                LOGGER.warn("Error occurred while committing metrics for schema records creation.", ex);
             }
         }
         LOGGER.warn(MessageFormat.format("Schema committer thread interrupted. {} metrics committed by this thread.", jobCounter.get()));
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeOnlySchemaRecord.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeOnlySchemaRecord.java
@@ -31,6 +31,8 @@
 
 package com.salesforce.dva.argus.entity;
 
+import com.google.common.base.Objects;
+
 import java.text.MessageFormat;
 
 /**
@@ -101,14 +103,7 @@ public class ScopeOnlySchemaRecord {
 
 		ScopeOnlySchemaRecord other = (ScopeOnlySchemaRecord) obj;
 
-		if (scope == null) {
-			if (other.scope != null) {
-				return false;
-			}
-		} else if (!scope.equals(other.scope)) {
-			return false;
-		}
-		return true;
+		return Objects.equal(scope, other.scope);
 	}
 
 	@Override
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
@@ -16,9 +16,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
-import com.google.common.hash.BloomFilter;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import com.salesforce.dva.argus.entity.KeywordQuery;
 import com.salesforce.dva.argus.entity.Metric;
 import com.salesforce.dva.argus.entity.MetricSchemaRecord;
@@ -102,6 +105,8 @@ public class ElasticSearchSchemaService extends AbstractSchemaService {
 	private HashAlgorithm _idgenHashAlgo;
 
 	private boolean _useScopeMetricNamesIndex;
+
+	private Cache<String, List<MetricSchemaRecord>> cacheResults;
 
 	@Inject
 	public ElasticSearchSchemaService(SystemConfiguration config, MonitorService monitorService) {
@@ -214,6 +219,12 @@ public class ElasticSearchSchemaService extends AbstractSchemaService {
 
 		_createIndexIfNotExists(SCOPE_AND_METRIC_INDEX_NAME, _replicationFactorForScopeAndMetricIndex,
 				_numShardsForScopeAndMetricIndex, () -> _createScopeAndMetricMappingsNode());
+
+		cacheResults = CacheBuilder.newBuilder()
+				.maximumSize(1000)
+				.recordStats()
+				.expireAfterAccess(10, TimeUnit.SECONDS)
+				.build();
 	}
 
 
@@ -509,11 +520,19 @@ public class ElasticSearchSchemaService extends AbstractSchemaService {
 		String queryJson = _constructTermAggregationQuery(query, type);
 		try {
 
-			_logger.debug("getUnique POST requestUrl {} queryJson {}", requestUrl, queryJson);
-			Response response = _esRestClient.performRequest(HttpMethod.POST.getName(), requestUrl, Collections.emptyMap(), new StringEntity(queryJson));
-			String str = extractResponse(response);
-			List<MetricSchemaRecord> records = SchemaService.constructMetricSchemaRecordsForType(
-					toEntity(str, new TypeReference<List<String>>() {}), type);
+			List<MetricSchemaRecord> records = cacheResults.get(queryJson, () -> {
+				try {
+					_logger.debug("getUnique POST requestUrl {} queryJson {}", requestUrl, queryJson);
+					Response response = _esRestClient.performRequest(HttpMethod.POST.getName(), requestUrl, Collections.emptyMap(), new StringEntity(queryJson));
+					String str = extractResponse(response);
+					return SchemaService.constructMetricSchemaRecordsForType(toEntity(str, new TypeReference<List<String>>() {}), type);
+				} catch (IOException e) {
+					throw new SystemException(e);
+				}
+			});
+
+			_logger.debug("getUnique POST CacheStats hitCount {} requestCount {} evictionCount {}",
+					cacheResults.stats().hitCount(), cacheResults.stats().requestCount(), cacheResults.stats().evictionCount());
 
 			if (query.isQueryOnlyOnScope() && RecordType.SCOPE.equals(type)) {
 				_monitorService.modifyCounter(Counter.SCOPENAMES_QUERY_COUNT, 1, tags);
@@ -538,8 +557,8 @@ public class ElasticSearchSchemaService extends AbstractSchemaService {
 			} else {
 				return records.subList(fromIndex, query.getLimit() * query.getPage());
 			}
-		} catch (IOException e) {
-			throw new SystemException(e);
+		} catch (ExecutionException e) {
+			throw new SystemException(e.getCause());
 		}
 	}
 
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
@@ -1428,7 +1428,7 @@ public class ElasticSearchSchemaService extends AbstractSchemaService {
 		 */
 		ELASTICSEARCH_INDEXING_BATCH_SIZE("service.property.schema.elasticsearch.indexing.batch.size", "10000"),
 
-		ELASTICSEARCH_USE_SCOPE_AND_METRIC_INDEX("service.property.schema.elasticsearch.use.scopeandmetric.index", "true"),
+		ELASTICSEARCH_USE_SCOPE_AND_METRIC_INDEX("service.property.schema.elasticsearch.use.scopeandmetric.index", "false"),
 
 		/** The hashing algorithm to use for generating document id. */
 		ELASTICSEARCH_IDGEN_HASH_ALGO("service.property.schema.elasticsearch.idgen.hash.algo", "MD5"),
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
@@ -547,6 +547,7 @@ public class ElasticSearchSchemaService extends AbstractSchemaService {
 			} else {
 				return records.subList(fromIndex, query.getLimit() * query.getPage());
 			}
+
 		} catch (IOException e) {
 			throw new SystemException(e);
 		}
@@ -923,8 +924,13 @@ public class ElasticSearchSchemaService extends AbstractSchemaService {
 				}
 
 				if (item.update != null && item.update.status != HttpStatus.SC_OK) {
-					_logger.warn("Failed to update document. Reason: " + new ObjectMapper().writeValueAsString(item.update.error));
-					failedIds.add(item.update._id);
+
+					if (item.update.status == HttpStatus.SC_CONFLICT) {
+						updateRequiredIds.add(item.update._id);
+					} else {
+						_logger.warn("Failed to update document. Reason: " + new ObjectMapper().writeValueAsString(item.update.error));
+						failedIds.add(item.update._id);
+					}
 				}
 			}
 		}
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
@@ -222,7 +222,6 @@ public class ElasticSearchSchemaService extends AbstractSchemaService {
 
 		cacheResults = CacheBuilder.newBuilder()
 				.maximumSize(1000)
-				.recordStats()
 				.expireAfterAccess(10, TimeUnit.SECONDS)
 				.build();
 	}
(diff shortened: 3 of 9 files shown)


ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -103,6 +106,8 @@
 
 	private boolean _useScopeMetricNamesIndex;
 
+	private Cache<String, List<MetricSchemaRecord>> cacheResults;
```
dongpujin: For each scroll, does the client request always hitting the same backend host? How about the case there initial request hit host A, which fetches data stored in host A's cache. Next scroll send request to host B, which does not have cache and result a miss. Can this happen?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -214,6 +219,11 @@ public Builder customizeRequestConfig(Builder requestConfigBuilder) {
 
 		_createIndexIfNotExists(SCOPE_AND_METRIC_INDEX_NAME, _replicationFactorForScopeAndMetricIndex,
 				_numShardsForScopeAndMetricIndex, () -> _createScopeAndMetricMappingsNode());
+
+		cacheResults = CacheBuilder.newBuilder()
```
dongpujin: nit: comply with existing var naming convention, _varname

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
naveenreddykarri: Yeah, I made that assumption that typically it will hit the same backend host. If not this solution wouldn't work well. Using Redis cache will be better. I will need to verify in QA cluster.

dilipdevaraj-sfdc: The web service calls go via load balancer so the scrolling request can goto any web server. I don't think local cache will be hit often since scrolling request may not end up on same web server.
We already have CachedDiscoveryService, where the results from elastic search are cached. Subsequent discovery service calls are fetched from this distributed Redis cache.

dongpujin: Do you know if the load balancer implements sticky session? If LB implements sticky session for some duration, this in memory cache approach would work.

dilipdevaraj-sfdc: No it uses JWT. Sticky session was not used so a bad actor does not bring a web server down.

ArgusClient/src/main/java/com/salesforce/dva/argus/client/SchemaCommitter.java
```diff
@@ -84,7 +84,7 @@ public void run() {
                 Thread.currentThread().interrupt();
                 break;
             } catch (Throwable ex) {
-                LOGGER.info("Error occurred while committing metrics for schema records creation.", ex);
+                LOGGER.warn("Error occurred while committing metrics for schema records creation.", ex);
```
rajsarkapally: You need {} to have error message as part of the log

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeOnlySchemaRecord.java
```diff
@@ -101,14 +103,7 @@ public boolean equals(Object obj) {
...
 
-		if (scope == null) {
-			if (other.scope != null) {
-				return false;
-			}
-		} else if (!scope.equals(other.scope)) {
-			return false;
-		}
-		return true;
+		return Objects.equal(scope, other.scope);
```
rajsarkapally: If two different scopeonlyschemarecords have same scope, should they be considered as equal?

ArgusClient/src/main/java/com/salesforce/dva/argus/client/SchemaCommitter.java
naveenreddykarri: It has void warn(String var1, Throwable var2); so we can pass an exception, which should log call stack + message.

naveenreddykarri: https://stackoverflow.com/questions/14606293/how-to-log-as-much-information-as-possible-for-a-java-exception

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/ScopeOnlySchemaRecord.java
nkunal: yes they should be considered equal

naveenreddykarri: yes, since for every metric on put we create a scopeonlyschemarecord if not present in bloomfilter, we want to check they are logically equivalent. The logic was already correct, I just simplified it.

dilipdevaraj-sfdc:
@naveenreddykarri 
I see there is some overlap in this PR and one by @rajsarkapally 
Can you take a look at https://github.com/salesforce/Argus/pull/847 and see how best to consolidate this or incorporate any work from that PR.

naveenreddykarri:
@dilip-devaraj. Yes, I looked at @rajsarkapally changes and added feedback. Both set of changes should be good to check-in.

dilipdevaraj-sfdc:
@naveenreddykarri 
Whats the difference in usecase between his and yours cts changes ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -1358,7 +1428,7 @@ public void setName(String name) {
 		 */
 		ELASTICSEARCH_INDEXING_BATCH_SIZE("service.property.schema.elasticsearch.indexing.batch.size", "10000"),
 
-		ELASTICSEARCH_USE_SCOPE_AND_METRIC_INDEX("service.property.schema.elasticsearch.use.scopeandmetric.index", "false"),
+		ELASTICSEARCH_USE_SCOPE_AND_METRIC_INDEX("service.property.schema.elasticsearch.use.scopeandmetric.index", "true"),
```
nkunal: we should keep the default as false, and enable it via config

naveenreddykarri: done

naveenreddykarri:
Hey @dilip-devaraj  His changes will enable deleting older documents in metadata_index. Mine will enable deleting older documents for scopenames, scopemetricnames indices.

dilipdevaraj-sfdc:
Wouldn't the deletion of ES documents still happen based on mts field? - ie schema that has not been modified more than X num days will be deleted.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -16,9 +16,13 @@
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
-import com.google.common.hash.BloomFilter;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.google.common.cache.Cache;
```
dilipdevaraj-sfdc: please remove unused imports

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -774,49 +784,56 @@ private void _upsertScopeAndMetrics(List<ScopeAndMetricOnlySchemaRecord> records
...
+				requestBody = _updateScopeAndMetricOnlyMapper.writeValueAsString(updateSchemaRecordList);
+				putResponse = _performRequest(requestUrl, requestBody);
+
+				failedResponses = _parseFailedResponses(putResponse);
+
+				// We collect new failures.
+				failedIds.addAll(failedResponses.getLeft());
+
+				// Do not collect update failures if they fail with 409 (version_conflict_engine_exception).
+				// failedIds.addAll(failedResponses.getRight());
```
dilipdevaraj-sfdc: Can unused code be removed , unless they are left intentionally.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -825,67 +842,120 @@ private void _upsertScopes(List<ScopeOnlySchemaRecord> records) {
...
+				requestBody = _updateScopeOnlyMapper.writeValueAsString(updateSchemaRecordList);
+				putResponse = _performRequest(requestUrl, requestBody);
+
+				failedResponses = _parseFailedResponses(putResponse);
+
+				// We collect new failures.
+				failedIds.addAll(failedResponses.getLeft());
+
+				// Do not collect update failures if they fail with 409 (version_conflict_engine_exception).
+				// failedIds.addAll(failedResponses.getRight());
```
dilipdevaraj-sfdc: same here

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -1440,6 +1510,7 @@ public void setItems(List<Item> items) {
 		static class Item {
 			private CreateItem create;
 			private CreateItem index;
```
dilipdevaraj-sfdc: Are we still using index anywhere?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -825,67 +842,120 @@ private void _upsertScopes(List<ScopeOnlySchemaRecord> records) {
...
+			ScopeOnlySchemaRecordList createSchemaRecordList = new ScopeOnlySchemaRecordList(records, _idgenHashAlgo);
+			String requestBody = _createScopeOnlyMapper.writeValueAsString(createSchemaRecordList);
+			PutResponse putResponse = _performRequest(requestUrl, requestBody);
+
+			Pair<List<String>, List<String>> failedResponses = _parseFailedResponses(putResponse);
+
+			List<String> failedIds = failedResponses.getLeft();
+			List<String> updateRequiredIds = failedResponses.getRight();
+
+			if (updateRequiredIds.size() > 0) {
```
dilipdevaraj-sfdc: Have we measured the performance of schema ingestion when we use create followed by update, versus using just index ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -774,49 +784,56 @@ private void _upsertScopeAndMetrics(List<ScopeAndMetricOnlySchemaRecord> records
...
 
-					if(item.index != null && item.index.status == HttpStatus.SC_NOT_FOUND) {
-						_logger.warn("Scope Index does not exist. Error: " + new ObjectMapper().writeValueAsString(item.index.error));
-						recordsToRemove.add(recordList.getRecord(item.index._id));
-					}
-				}
-				if(recordsToRemove.size() != 0) {
-					_logger.info("{} records were not written to ES", recordsToRemove.size());
-					records.removeAll(recordsToRemove);
+			if (updateRequiredIds.size() > 0) {
```
dilipdevaraj-sfdc: Here we are always making 2 ES calls one for create and another for update
After some time of running schema clients ES would have already seen that schema. If we now try to insert say 10000 records in one batch
- so in first call "create" will always fail and populate the entire 10000 record ids in updateRequiredIds
- now we do the regular "update" operation, but now instead of having a smaller subset of records it will have the entire 10000 records.
Is there any way we can optimize this ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/SchemaRecordList.java
```diff
@@ -15,6 +16,27 @@
...
 public class SchemaRecordList {
 
+    public static void addCreateJson(JsonGenerator jgen, String id, String fieldsData) throws IOException {
+
+        jgen.writeRaw("{ \"create\" : {\"_id\" : \"" + id + "\"}}");
+        jgen.writeRaw(System.lineSeparator());
+
+        long currentTimeMillis = System.currentTimeMillis();
+
+        String createTimeStampField = "\"cts\":" + currentTimeMillis;
```
dilipdevaraj-sfdc: What is the real usecase/benefit of having the cts field versus performance impact ?
- Right now we are already leveraging the mts field when we want to delete older records
- We also using "mts field" from ES index1 to aggregate index to detect highest number of incoming metrics - we use Kibana for this and can go back a few days to detect bad actors sending too many new metrics ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -774,49 +784,56 @@ private void _upsertScopeAndMetrics(List<ScopeAndMetricOnlySchemaRecord> records
...
+				ScopeAndMetricOnlySchemaRecordList updateSchemaRecordList = new ScopeAndMetricOnlySchemaRecordList(updateRequiredRecords, _idgenHashAlgo);
+				requestBody = _updateScopeAndMetricOnlyMapper.writeValueAsString(updateSchemaRecordList);
+				putResponse = _performRequest(requestUrl, requestBody);
+
+				failedResponses = _parseFailedResponses(putResponse);
+
+				// We collect new failures.
+				failedIds.addAll(failedResponses.getLeft());
+
+				// Do not collect update failures if they fail with 409 (version_conflict_engine_exception).
```
dilipdevaraj-sfdc: Can you add comment in code
"This usually happens when there is another concurrent update happening to mts field"

dilipdevaraj-sfdc:
This needs to be perf tested, with lots of documents to see how it behaves when there are conflicts, and multiple records need to be written as part of update when all records for create have failed.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/SchemaRecordList.java
naveenreddykarri: adding kunal's response.

Hi Dilip, following up on some of the questions you had about cts field on Naveen's PR. 
As you know the 'mts' field is populated so that we can leverage it for deleting older docs. The aggs job for cardinality detection and reporting uses mts as the indicator of time. Since we refresh indexes every 24 hours, that is not an accurate indication of new time series creation
Hence we want to send a 'cts' field for all new docs (via the create ES operation). And we will change the aggs job to use the cts field for that. Once this is done then we will have an accurate view of "who created new timeseries in the last N minutes.
as far as performance is considered, that is not an issue, since we are protectiing this with a bloomfilter. And also another datapoint, ES can take in even more traffic than what is doing now. Recent example was last week, when Colby was running both AKC and schmea clients in parallel, so we are hitting ES with double the traffic

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
naveenreddykarri: fixed.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
naveenreddykarri: done

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
naveenreddykarri: had left it intentionally but will remove as suggested.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
naveenreddykarri: we are using bloomfilter to reduce the number of create/update operations. One option could be to try update first and if fails do a create. I am not sure if that would lead to overall better perf though.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
naveenreddykarri: done

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
naveenreddykarri: will remove this later. It can be removed once both I and raj check-in our changes.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
naveenreddykarri: 8457 new metrics were indexed in 3024 ms. [Just index]
4984 new scope and metric names were indexed in 3926 ms. [create / update]
Added some logging that shows the time spent.

naveenreddykarri:
8457 new metrics were indexed in 3024 ms. [Just index]
4984 new scope and metric names were indexed in 3926 ms. [create / update]
Added some logging that shows the time spent.

naveenreddykarri:
Tests will send a separate PR. I want to get Raj's changes and then add mine as it shows changes to the complete ElasticSearchSchemaServiceTest file.

PR #862, opened 2018-08-02
Fill Transform to round to the current minute

studanshu:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/FillTransform.java
```diff
@@ -146,19 +148,9 @@ private static long _parseTimeIntervalInSeconds(String interval) {
         long offsetInSeconds = _parseTimeIntervalInSeconds(constants.get(3));
         double value = Double.parseDouble(constants.get(4));
 
-        SystemAssert.requireArgument(startTimestamp < endTimestamp, "End time must occure later than start time!");
+        SystemAssert.requireArgument(startTimestamp < endTimestamp, "End time must occur later than start time!");
         SystemAssert.requireArgument(windowSizeInSeconds >= 0, "Window size must be greater than ZERO!");
 
-        boolean isDivisible = ((startTimestamp - endTimestamp) % (windowSizeInSeconds * 1000)) == 0;
```
sundeepsf: we still need this snapping logic if the window is not divisible by the step size

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/FillTransform.java
```diff
@@ -173,6 +180,12 @@ private static long _parseTimeIntervalInSeconds(String interval) {
         return lineMetrics;
     }
 
+    private long roundOffToNearestMinute(long timeInMillis) {
```
sundeepsf: minor, this is not nearest minute, but the current minute

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/FillTransform.java
```diff
@@ -181,20 +180,30 @@ private static long _parseTimeIntervalInSeconds(String interval) {
...
+                    retVal = Long.parseLong(timeStr);
+                }
+            } catch (NumberFormatException nfe) {
+                throw new SystemException("Could not parse time.", nfe);
             }
-            return Long.parseLong(timeStr);
-        } catch (NumberFormatException nfe) {
-            throw new SystemException("Could not parse time.", nfe);
         }
+        return roundOffToNearestMinute(retVal);
```
sundeepsf: actually thinking a bit more, can we do this only if the step size is more than a minute

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/FillTransform.java
studanshu: @sundeepsf will make changes.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/FillTransform.java
studanshu: ok.

PR #877, opened 2018-08-14
1) Add metrics for bloom filter 2) Use schema committer hostname hash…

dilipdevaraj-sfdc:
… so don't need a separate config

### DIFF
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
@@ -1,5 +1,7 @@
 package com.salesforce.dva.argus.service.schema;
 
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -373,11 +388,12 @@ public void run() {
 		}
 
 		private void _checkBloomFilterUsage() {
-			_logger.info("Metrics Bloom approx no. elements = {}", bloomFilter.approximateElementCount());
+			_monitorService.modifyCounter(MonitorService.Counter.BLOOMFILTER_APPROXIMATE_ELEMENT_COUNT, bloomFilter.approximateElementCount(), null);
```
naveenreddykarri: This method is run once every 10 mins I think. If it is a counter to be useful, should it run atleast once a minute? May be we then don't do the logging?

dilipdevaraj-sfdc: Yes its fine to run it every 10 minutes, and not every minute.

PR #878, opened 2018-08-15
Adding case when expression starts with relative time

studanshu:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/notifier/DefaultNotifier.java
```diff
@@ -106,17 +109,22 @@ public void sendNotification(NotificationContext notificationContext) {
...
+                String timeStr = m.group();
+                Long absoluteTime = MetricReader.getTime(relativeTo, timeStr.substring(1));
+                expression = expression.replace(timeStr, ("" + timeStr.charAt(0)) + absoluteTime);
+            }
+            absoluteExpression = expression.substring(1);
+        } catch (Exception ex) {
+             _logger.error("Exception occurred while converting expression to absolute time.", ex.getMessage());
         }
-        return expression;
+        return absoluteExpression;
```
sundeepsf: just to make sure, does the upstream code still work ok if you return empty string in case of exception?

sundeepsf: I think its better to leave the absolute expression out of the notfication message if its empty. It will only confuse the users if we include "" or relative expression

studanshu: @sundeepsf made changes

sundeepsf: Like discussed, please make these 2 changes -

1. Remove the triggering event timestamp
2. If the absolute expression is "", show relative expression in "evaluated metric expression" field

PR #879, opened 2018-08-15
Made changes to add extra params to downsample transform to be able to generate absolute intervals

sundeepsf:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/DownsampleTransform.java
```diff
@@ -55,148 +55,188 @@
...
+
+		String[] expArr = constants.get(0).split("-");
+
+		SystemAssert.requireArgument(expArr.length == 2, "This downsampler expression need both unit and type.");
+
+		Long startTime = null;
+		Long endTime = null;
+		Double defaultValue = null;
+		if(constants.size()==4) {
+			long currMinuteStartTime = 60*1000*(System.currentTimeMillis()/(60*1000));
```
dilipdevaraj-sfdc: Instead of asking the user to enter additional constants, why can't we do
DOWNSAMPLE(<time_series>]*,#downsampler#)
DOWNSAMPLE_QUERYTIME(<time_series>]*,#downsampler#)

sundeepsf: We can't get the query time easily as the underlying timeseries can itself be wrapped in multiple transforms. I did explore that option. If you know of a better way to get those params, let me know

dilipdevaraj-sfdc: May be we can discuss by looking at code in MetricReader
Right now we are basing it off current timestamp but if user were to pass relative timestamp like  -100m:-10m:s:m
compared to
-110m:-10m:s:m

Also were you able to manually test the added functionality in transform ?

sundeepsf: Logic is based on relative timestamp only. I am testing it in qa right now

sundeepsf: Tested on our dev cluster. Seems to be working fine

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/DownsampleTransform.java
```diff
@@ -55,148 +55,188 @@
...
+
+		String[] expArr = constants.get(0).split("-");
+
+		SystemAssert.requireArgument(expArr.length == 2, "This downsampler expression need both unit and type.");
+
+		Long startTime = null;
+		Long endTime = null;
+		Double defaultValue = null;
+		if(constants.size()==4) {
+			long currMinuteEndTime = 60*1000*(System.currentTimeMillis()/(60*1000)) + 60*1000;
```
dilipdevaraj-sfdc: Can you add java comment why we are basing  time as end of current minute

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/DownsampleTransform.java
```diff
@@ -55,148 +55,188 @@
...
+			startTime = MetricReader.getTime(currMinuteEndTime, constants.get(1));
+			endTime = MetricReader.getTime(currMinuteEndTime, constants.get(2));
+			defaultValue = Double.parseDouble(constants.get(3));
```
dilipdevaraj-sfdc: Can you add java comment of use of defaultValue (maybe this is acting as a a FILL value), since DOWNSAMPLE based on user query time would only need extra startTime and endTime as additional parameters.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/DownsampleTransform.java
```diff
@@ -55,148 +55,188 @@
...
+
+	//~ Methods **************************************************************************************************************************************
+
+	@Override
+	public List<Metric> transform(List<Metric> metrics) {
+		throw new UnsupportedOperationException("Downsample transform need constant input!");
+	}
+
+	@Override
+	public List<Metric> transform(List<Metric> metrics, List<String> constants) {
```
dilipdevaraj-sfdc: So now the DOWNSAMPLE based of user query time would need to be explicitly specified twice like 
DOWSAMPLE(-100m:-10m:scope:metric:sum:10m-sum, #10m-sum#,#-100m#, #-10m#)
Is that right ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/DownsampleTransform.java
sundeepsf: will do

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/DownsampleTransform.java
sundeepsf: yes, also you need to specify the default value. Like this -

DOWNSAMPLE(-1h:argus.core:alerts.scheduled:zimsum, #5m-sum#, #-1h#, #-0m#, #0#)

dilipdevaraj-sfdc: Does this also work for epoch timestamps ?
DOWNSAMPLE(1534377927:1534377928:argus.core:alerts.scheduled:zimsum, #5m-sum#, #1534377927#, #1534377928#, #0#)
and 
DOWNSAMPLE(1534377927000:1534377928000:argus.core:alerts.scheduled:zimsum, #5m-sum#, #1534377927000#, #1534377928000#, #0#)

Since some users specify epoch timestamp in tsdb query, if they want to lookup an incident occurence at a later point of time

sundeepsf: Yes, it does. I am reusing the parser code from MetricReader.java which seems to automatically handle this as well

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/DownsampleTransform.java
```diff
@@ -206,47 +247,47 @@ private long getWindowStartTime(long time, String windowUnit, long windowSize){
...
+		result=time/1000;
+		switch(field){
+		case Calendar.SECOND:
+			result=result-(result%secondOffset);
+			break;
+		case Calendar.MINUTE:
+			result=result-(result%minuteOffset);
+			break;
+		case Calendar.HOUR:
+		case Calendar.HOUR_OF_DAY:
```
dilipdevaraj-sfdc: In AnnotationReader.jj we support DAY as well
Can we support Calendar.DAY here too ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/DownsampleTransform.java
```diff
@@ -205,7 +206,7 @@ public static Long downsamplerTimestamp(Long millitimestamp, long windowSize) {
 				startTime -= timeDrift;
 				endTime -= timeDrift;
 			}
-			windowStart = startTime;
+			windowStart = getWindowStartTime(startTime,windowUnit,windowSize);
```
dilipdevaraj-sfdc: Pls add space after comma

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/DownsampleTransform.java
sundeepsf: This is existing logic, did not want to touch that with this PR

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/DownsampleTransform.java
sundeepsf: done

PR #880, opened 2018-08-16
Adding cts field and updating mts field for metadata_index

rajsarkapally:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -743,17 +747,32 @@ protected void upsert(List<MetricSchemaRecord> records) {
...
-						_logger.warn("Index does not exist. Error: " + new ObjectMapper().writeValueAsString(item.index.error));
-						recordsToRemove.add(msrList.getRecord(item.index._id));
+				}
+				if(updateMtsFieldList.size()>0) {
+					_logger.debug("mts filed will be updated for docs with ids {}", updateMtsFieldList); 
+					Response response = updateMtsField(updateMtsFieldList,INDEX_NAME,TYPE_NAME);
+					PutResponse updateResponse = new ObjectMapper().readValue(extractResponse(response), PutResponse.class);
+					for(Item item: updateResponse.items) {
+						if(item.update != null && item.update.status != HttpStatus.SC_OK) {
+							_logger.debug("Failed to update mts field for metric {}. Reason: {}",msrList.getRecord(item.update._id),
```
naveenreddykarri: if(item.update.status == HttpStatus.SC_CONFLICT) {
we can do _logger.debug since it is expected
else _logger.warn so that we know any unexpected error.
only on _logger.warn we do recordsToRemove.add(msrList.getRecord(item.update._id)); since in first case it means the update happened concurrently, so thinking we probably don't want to count it as failure here so that 
we don't see it at the end for concurrent update failures.
_logger.warn("{} records were not written to ES", recordsToRemove.size());

any other failure we should see the warning.

rajsarkapally: @naveenreddykarri Agree..will change to warn.. But when item.update.status == HttpStatus.SC_CONFLICT meaning the mts field of the record is not updated. If we do not write this record to the bloom filter, when the same record comes again, mts field will be updated.

naveenreddykarri:
LGTM

rajsarkapally:
@naveenreddykarri As you said, lets do spacing/formatting in a different CR.  Rolled back all spacing and formatting
I have changed logger to warn

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -743,17 +747,32 @@ protected void upsert(List<MetricSchemaRecord> records) {
...
 				for(Item item : putResponse.items) {
-					if(item.create != null && item.create.status != HttpStatus.SC_CONFLICT && item.create.status != HttpStatus.SC_CREATED) {
-						_logger.warn("Failed to index metric. Reason: " + new ObjectMapper().writeValueAsString(item.create.error));
-						recordsToRemove.add(msrList.getRecord(item.create._id));
+					if(item.create != null) {
+						if(item.create.status == HttpStatus.SC_CONFLICT) {
+							updateMtsFieldList.add(item.create._id);
+						}else if(item.create.status != HttpStatus.SC_CREATED) {
+							_logger.warn("Failed to index metric {}. Reason: {}", msrList.getRecord(item.create._id),
+									new ObjectMapper().writeValueAsString(item.create.error));
```
nkunal: For each failed response, if we log this msg, it flood the logs and can bring down entire processing.

naveenreddykarri:
LGTM

PR #881, opened 2018-08-16
Long support 2

adelaidechambers:
Add support for long metric values and long trigger threshold values in Argus.

Note: This is essentially a copy of a previous PR to resolve CLA issue. I've copied comments from the previous PR over to keep track of them.

@colbyguan @pratiksha-shah @kgowdru @sundeepsf @rajsarkapally @dilipdevaraj-sfdc @nkunal 

From the earlier PR:
> Allow users to input metric data with long data values to Argus and long alert trigger threshold values. This way large values can be maintained precisely and triggers can be appropriately fired based off of these larger values.


adelaidechambers:
Kunal:
> @kgowdru @adelaidechambers This is a fairly large change. We can continue with the review of this PR and incorporate feedback, but this should not get merged (and released) in our next release. Lets use this PR as an opportunity to come up with a set of integration/end-to-end/regression testing framework for Argus in general

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Trigger.java
```diff
@@ -88,22 +90,43 @@
...
 		@Override
 		public void serialize(Trigger trigger, JsonGenerator jgen, SerializerProvider provider) throws IOException, JsonProcessingException {
-			
 			jgen.writeStartObject();
 			
 			jgen.writeStringField("id", trigger.getId().toString());
 			jgen.writeStringField("name", trigger.getName());
 			jgen.writeStringField("type", trigger.getType().name());
-			jgen.writeNumberField("threshold", trigger.getThreshold().doubleValue());
+
```
adelaidechambers: > Important serialization/deserialization code for Triggers

adelaidechambers:
Dilip:
> @adelaidechambers
> pls @mention reviewers so they can provide any review comments.

adelaidechambers:
Dilip:
> @nkunal @colbyguan @sundeepsf @rajsarkapally @asinghsfdc
> Can you also pls review this PR, so @adelaidechambers can have sufficient time to incorporate feedback and do necessary testing.
> We can then make a call of how and when we want to deploy this feature change.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/NumberOperations.java
```diff
@@ -0,0 +1,879 @@
+package com.salesforce.dva.argus.entity;
```
adelaidechambers: This is an important class, centralizes all of the mathematical/logical operations on Numbers.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/DiscoveryService.java
```diff
@@ -49,7 +49,7 @@
 	
 	/** This should be a configuration. For now, this is how we reached on a value of 2M. 
 	 *  A datapoint in Argus is a tuple containing a Long timestamp (8 bytes with some additional Java Wrapper Class bytes) 
-	 *  and a Double value (8 bytes with some additional Java Wrapper Class bytes). We would then consider a datapoint to 
+	 *  and a Number value (8 bytes with some additional Java Wrapper Class bytes). We would then consider a datapoint to 
 	 *  take up around 40 bytes of memory.
```
adelaidechambers: > dilipdevaraj-sfdc 29 days ago  Member
> When we are storing it as Number is the comment still true - that value portion of datapoint will be 8 bytes ?
> 
>    
> adelaidechambers 28 days ago  Member
> I think this is still true. From what I've found, the data value will still just be the size of a long/double, which is 8 bytes, plus the class wrapper.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/AbsoluteValueMapping.java
```diff
@@ -46,19 +48,17 @@
...
-            absDatapoints.put(entry.getKey(), absValue);
-        }
-        return absDatapoints;
+    public Map<Long, Number> mapping(Map<Long, Number> originalDatapoints) {
+    	Map<Long, Number> absDatapoints = new HashMap<>();
+    	
+    	for (Entry<Long, Number> entry : originalDatapoints.entrySet()) {
+    		absDatapoints.put(entry.getKey(), NumberOperations.getAbsValue(entry.getValue()));
+    	}
+    	return absDatapoints;
```
adelaidechambers: > colbyguan 6 days ago  Contributor
> Change from 4-space to 2-space indent seems to be crowding the diff; can we set it back?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/CountTransformWrapUnion.java
```diff
@@ -75,20 +76,20 @@
         return Arrays.asList(newMetric);
     }
     
-    private Map<Long, Double> _collate(List<Metric> metrics) {
-        Map<Long, Double> collated = new HashMap<>();
-
+    private Map<Long, Number> _collate(List<Metric> metrics) {
+        Map<Long, Number> collated = new HashMap<>();
```
adelaidechambers: > colbyguan 6 days ago  Contributor
> Same tabbing thing here

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/CountValueUnionReducer.java
```diff
@@ -42,17 +44,17 @@
...
     //~ Methods **************************************************************************************************************************************
-
+    
     @Override
-    public Double reduce(List<Double> values) {
-        Double sum = 0.0;
-
-        for (Double value : values) {
+    public Number reduce(List<Number> values) {
+        Number sum = 0;
```
adelaidechambers: > colbyguan 6 days ago  Contributor
> And here

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/DivideValueZipper.java
```diff
@@ -46,17 +47,16 @@
...
-            Double original = (originalDp == null) ? 0.0 : originalDp;
-            Double base = (baseDp == null) ? 1.0 : baseDp;
-
-            SystemAssert.requireArgument(base != 0.0, "Datapoints in base metric shouldn't contain zero!");
-            
-            return (original / base);
-        } catch (Exception e) {
-            throw new SystemException("Fail to parse the double value of original Datapoint or base Datapoint!", e);
-        }
+    public Number zip(Number originalDp, Number baseDp) {
```
adelaidechambers: > colbyguan 6 days ago  Contributor
> Same indent thing here

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/ShiftValueMapping.java
```diff
@@ -48,22 +48,26 @@
...
 
-        for (Entry<Long, Double> entry : originalDatapoints.entrySet()) {
+        for (Entry<Long, Number> entry : originalDatapoints.entrySet()) {
+        	if (offset >= 0) {
+        		SystemAssert.requireArgument((entry.getKey() <= Long.MAX_VALUE - offset && entry.getKey() >= Long.MIN_VALUE), 
+        				"You are not allowed to shift like this, be nice to me!");
+        	} else {
+        		SystemAssert.requireArgument((entry.getKey() <= Long.MAX_VALUE && entry.getKey() >= Long.MIN_VALUE - offset),
+        				"You are not allowed to shift like this, be nice to me!");
+        	}
```
adelaidechambers: > colbyguan 6 days ago  Contributor
> This extra if-else appears to do the same thing that the old code does, with an extra bounding of the entry original timestamp to being between valid Long values (which, if violated, would have been a user input error anyway). Was there a certain reason for this change?
> 
> if offset positive or 0, then ensure key+offset <= Long.MAX
> if offset negative, then ensure key+offset >= Long.Min
>    
> adelaidechambers 3 days ago  Member
> The old behavior was a NOP. The statement 'if offset positive or 0 then ensure key+offset <= Long.MAX' will always be true here. Key and offset are both treated as longs and summed into the long newTimestamp, so is guaranteed to be a long regardless of whether that long value has overflowed. Assuming offset is a valid long, though, and calculating key <= Long.MAX - offset could be violated.
> For example, consider the following: let key = Long.MAX_VALUE, offset = 1. Then, key + offset = (Long.MAX_VALUE + 1) = Long.MIN_VALUE <= Long.MAX_VALUE. But, in the new check, key = Long.MAX_VALUE > Long.MAX_VALUE - 1 = Long.MAX_VALUE - offset. So,, this wouldn't cause an exception in the old loop but would in the new one. I assume those kinds of shifts were the purpose of the exception, so now it will catch them.
> I added the check that the constants were within the bounds for a long value first since, in theory, they could be double values outside of the scope of a long value. I agree that these exceptions would be caused by user error, but thought that was largely the intention of the original exception anyways (similar to providing an illegal constant in the other transforms).

ArgusSDK/src/test/java/com/salesforce/dva/argus/sdk/MetricServiceTest.java
```diff
@@ -69,14 +69,34 @@ public void testGetMetrics() throws IOException, TokenExpiredException {
             assertEquals(expected, result);
         }
     }
+
```
adelaidechambers: > dilipdevaraj-sfdc 29 days ago  • 
> edited 
>  Member
> Do we have tests which validates the intended feature change, something like
> Case 1
> // datapoint as long
> putDatapoints(timestamp, 101)
> getDatapoints
> // assert value is exact precision as long 101 and not 101.0
> 
> similarly
> Case2
> // datapoint as double
> putDatapoints(timestamp, 101.1)
> getDatapoints
> // assert value is exact precision as double 101.1 and not 101
> 
> Case3
> Long number with greater than 53 bits (ie greater than 2^(53)-1, or 9007199254740991 )- so precision is not lost as with old implementation, and returned precision should be exact.
> 
>    
> adelaidechambers 28 days ago  Member
> I've run all of these tests manually, but I can add them to this as well.
> 
>    
> dilipdevaraj-sfdc 28 days ago  Member
> Its always a good practice to write unit tests if possible in addition to just manual tests. That way when there is a future release a new feature does not cause regression, since this will be captured via unit test.

ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/dto/EntityDTO.java
```diff
@@ -71,9 +75,14 @@
      */
     public static <D extends EntityDTO, E extends JPAEntity> D createDtoObject(Class<D> clazz, E entity) {
         D result = null;
+
```
adelaidechambers: > adelaidechambers 28 days ago  Member
> This works around a known bug for BigDecimal where BeanUtils throws an exception for any null values.

colbyguan:
@adelaidechambers BTW I think we should hold off on merging master into this branch until the very end, so as to not crowd the commits / introduce other authors into the PR.

PR #886, opened 2018-08-20
Inertia and Subscription Validation for alerts

studanshu:


ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Trigger.java
```diff
@@ -480,5 +490,41 @@ public String value() {
         }
 
     }
+
+    private Long getMaximumIntervalLength(String expression) {
+        expression = "@" + expression.replaceAll("[\\s\\t\\r\\n\\f]*", "");
```
sundeepsf: seems like this extracting logic is very similar to the logic for computing absolute expressions. Maybe you can move the common logic to some kind of util class?

studanshu: Yes. I also thought so. But regexes are a bit different in both the cases.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Notification.java
```diff
@@ -431,6 +434,16 @@ public void setNotifierName(String notifierName) {
      */
     public void setSubscriptions(List<String> subscriptions) {
         this.subscriptions.clear();
+
+        for(String currentSubscription: subscriptions) {
+            if (this.getNotifierName().equals(AlertService.SupportedNotifier.GUS.getName())) {
+                if (currentSubscription.length() < 10)
+                    throw new IllegalArgumentException("GUS subjectId is incorrect.");
```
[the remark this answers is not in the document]
studanshu: https://github.com/salesforce/Argus/blob/develop/ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/notifier/GusNotifier.java#L199
We specify subjectId over here.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Trigger.java
sundeepsf: Seems this logic will be useful in other places also. This function seems out of place in trigger class. Could you move it to a util class?

studanshu: Ok.

PR #887, opened 2018-08-20
 Add ag directives with autocomplete features for scope, metric, tag and namespace

prestonfff:
Use the directive like: \<ag-scope-name default="scope" /\>, it will show a dropdown with autocomplete suggestions based on user input.


xizi-xu:
How can those fields be used in ag-metric?

cnardi-dev:
Any updates to wiki docs for use of fields with ag-metric?

PR #888, opened 2018-08-21
Use write VIP and clean up write path

rajsarkapally:


rajsarkapally:
@dilip-devaraj  Atul asked to get rid of using host names and remove iterator and use VIP instead. 
https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000005CMlZIAW/view

dilipdevaraj-sfdc:
@rajsarkapally  we should probably not be posting gus links here.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/tsdb/AbstractTSDBService.java
```diff
@@ -664,7 +609,7 @@ protected HttpResponse executeHttpRequest(HttpMethod requestType, String url,  C
 		/** The TSDB read endpoint. */
 		TSD_ENDPOINT_READ("service.property.tsdb.endpoint.read", "http://localhost:4466,http://localhost:4467"),
 		/** The TSDB write endpoint. */
-		TSD_ENDPOINT_WRITE("service.property.tsdb.endpoint.write", "http://localhost:4477,http://localhost:4488"),
+		TSD_ENDPOINT_WRITE("service.property.tsdb.endpoint.write", "http://argusdev-tsdb.data.sfdc.net:4466"),
```
dilipdevaraj-sfdc: pls change this to localhost

ArgusCore/src/main/java/com/salesforce/dva/argus/service/tsdb/AbstractTSDBService.java
```diff
@@ -112,18 +109,14 @@
 	private final ObjectMapper _mapper;
 	protected final Logger _logger = LoggerFactory.getLogger(getClass());
 
-	private final String[] _writeEndpoints;
+	private final String _writeEndpoint;
```
dilipdevaraj-sfdc: @asinghsfdc 
For write endpoint open source folks will probably not use it as VIP1, VIP2
instead they will most likely use  a hardware/software load balancer and use a singular VIP endpoint. (This new code only supports a singular VIP endpoint)

But the open source folks who don't have a load balancer configured can still leverage the same behaviour by using host1,host2....... via configuration 
(This is what older code supported, but will not work with this new change)

I think it makes sense  for the user configuration to accept multiple endpoints.

PR #889, opened 2018-08-22
Max datapoints

adelaidechambers:
Includes the max datapoints feature that will filter a metric's datapoints down to a given number of datapoints after querying and performing any transforms.

This PR is based off of the Long Support 2 PR, so most of the changes shown here are actually changes in that PR. Relevant changes in this PR are commented.

@colbyguan @kgowdru @sundeepsf @pratiksha-shah @rajsarkapally @dilipdevaraj-sfdc @nkunal


ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/MaxDatapointTransform.java
```diff
@@ -0,0 +1,286 @@
+package com.salesforce.dva.argus.service.metric.transform;
```
adelaidechambers: This class is entirely new. It is the transform that handles the filtering down to a maximum number of datapoints with different possible filtering algorithms.

ArgusCore/src/test/java/com/salesforce/dva/argus/service/metric/transform/MaxDatapointTransformTest.java
```diff
@@ -0,0 +1,621 @@
+package com.salesforce.dva.argus.service.metric.transform;
```
adelaidechambers: This class is entirely new. This tests the new Max Datapoint Transform.

ArgusCore/src/main/javacc/MetricReader.jj
```diff
@@ -241,6 +241,7 @@ TOKEN : { < ANOMALY_RPCA : "ANOMALY_RPCA" > }
 TOKEN : { < GROUPBY : "GROUPBY" > }
```
adelaidechambers: This class has been updated to support the MAX_DATAPOINT transform.

PR #891, opened 2018-08-24
Max datapoint updated

adelaidechambers:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/MaxDatapointTransform.java
```diff
@@ -0,0 +1,286 @@
+package com.salesforce.dva.argus.service.metric.transform;
```
adelaidechambers: This file is new. It is the transform that performs all of the filtering logic.

ArgusCore/src/test/java/com/salesforce/dva/argus/service/metric/transform/MaxDatapointTransformTest.java
```diff
@@ -0,0 +1,621 @@
+package com.salesforce.dva.argus.service.metric.transform;
```
adelaidechambers: This class is new. It tests Max Datapoint Transform, which performs all of the filtering logic.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/TransformFactory.java
```diff
@@ -190,6 +190,8 @@ public Transform getTransform(String functionName) {
                 return new AnomalyDetectionRPCATransform();
             case INTERPOLATE:
                 return new InterpolateTransform();
+            case MAX_DATAPOINT:
```
adelaidechambers: I've added the Max_Datapoint transform here.

ArgusCore/src/main/javacc/MetricReader.jj
```diff
@@ -241,6 +241,7 @@ TOKEN : { < ANOMALY_RPCA : "ANOMALY_RPCA" > }
 TOKEN : { < GROUPBY : "GROUPBY" > }
 TOKEN : { < GROUPBYTAG : "GROUPBYTAG" > }
 TOKEN : { < INTERPOLATE : "INTERPOLATE" > }
+TOKEN : { < MAX_DATAPOINT: "MAX_DATAPOINT" > }
```
adelaidechambers: I've added Max_Datapoint parsing logic here.

PR #897, opened 2018-08-30
Templatizing components of alerts

studanshu:
Using Freemarker library for easy templatization of alert, trigger, components. Various use cases specified in test file.


ArgusCore/src/main/java/com/salesforce/dva/argus/util/TemplateReplacement.java
```diff
@@ -0,0 +1,89 @@
...
+            templateString = singletonInstance.replaceKeywordsToLowerCase(templateString);
+            Template configuredTemplate = null;
+            try {
+                configuredTemplate = new Template("configuredTemplate", new StringReader(templateString), cfg);
+                StringWriter stringWriter = new StringWriter();
+                configuredTemplate.process(root, stringWriter);
+                generatedString = stringWriter.toString();
+            } catch (Exception e) {
+                _logger.error(MessageFormat.format("Exception occurred while applying template change on {0}, with error message {1}.", templateString, e.getMessage()));
+                return null;
```
sundeepsf: do you want to just return the original string in this case?

ArgusCore/src/main/java/com/salesforce/dva/argus/util/TemplateReplacement.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.io.*;
+import java.text.MessageFormat;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class TemplateReplacement {
```
sundeepsf: Better name is TemplateReplacer

ArgusCore/src/main/java/com/salesforce/dva/argus/util/TemplateReplacement.java
```diff
@@ -0,0 +1,89 @@
...
+        String lowercasedString = templateString;
+        Matcher m = Pattern.compile("(?i)\\$\\{scope\\}|\\$\\{metric\\}|\\$\\{tag\\..*?\\}|\\sscope\\s|\\smetric\\s|\\stag\\..*?\\s").matcher(templateString);
+        while (m.find()) {
+            String currentSubstring = m.group();
+            lowercasedString = lowercasedString.replace(currentSubstring, currentSubstring.toLowerCase());
+        }
+        return lowercasedString;
+    }
+
+    public static String applyTemplateChanges(DefaultAlertService.NotificationContext context, String templateString) {
```
sundeepsf: why this method needs to be static?

ArgusCore/src/main/java/com/salesforce/dva/argus/util/TemplateReplacement.java
```diff
@@ -0,0 +1,89 @@
...
+    private TemplateReplacement() {
+        cfg = new Configuration(Configuration.VERSION_2_3_28);
+        cfg.setDefaultEncoding("UTF-8");
+        cfg.setTemplateExceptionHandler(TemplateExceptionHandler.RETHROW_HANDLER);
+        cfg.setLogTemplateExceptions(false);
+        cfg.setWrapUncheckedExceptions(true);
+    }
+
+    public static TemplateReplacement getInstance( ) {
+        return singletonInstance;
```
sundeepsf: does this needs to be a singleton?Generally its not a good practice to use the singletons. In the constructor you can just check if the static variables are null and the initialize them if they are

ArgusCore/src/main/java/com/salesforce/dva/argus/util/TemplateReplacement.java
```diff
@@ -0,0 +1,89 @@
...
+
+public class TemplateReplacement {
+
+    private static final Logger _logger = LoggerFactory.getLogger(TemplateReplacement.class);
+    private static TemplateReplacement singletonInstance = new TemplateReplacement();
+    private static final String DATE_FORMAT = "MM/dd/yyyy HH:mm:ss z";
+    private static Configuration cfg;
+
+    private TemplateReplacement() {
+        cfg = new Configuration(Configuration.VERSION_2_3_28);
```
sundeepsf: you can initialize this stuff in a static block instead of making this class a singleton

ArgusCore/src/main/java/com/salesforce/dva/argus/util/TemplateReplacer.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class TemplateReplacer {
+
+    private static final Logger _logger = LoggerFactory.getLogger(TemplateReplacer.class);
+    private static TemplateReplacer singletonInstance = new TemplateReplacer();
```
sundeepsf: this instance is not required. You can initialize cfg in a static init block and make all the methods static

PR #905, opened 2018-09-05
Options to disable MQService producer or consumer

colbyguan:
Some ArgusCore-dependent modules (eg. ArgusClient) may not need the specific producer or consumer component of MQService. Before, even if the `Producer.java` class was unused, it would initiate a KafkaProducer connection to brokers; this prevents that if disable option is set to true


colbyguan:
@rajsarkapally handoff

colbyguan:
Last commit was a rebase. Tested in QA environment with webservices and ArgusClient working normally; merging

PR #908, opened 2018-09-06
Ability to push optional Metatags in a metric to a ES index

nkunal:
@naveenreddykarri please review


ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Metric.java
```diff
@@ -67,6 +67,7 @@
 	private String _units;
 	private final Map<Long, Double> _datapoints;
 	private MetricQuery _query;
+        private MetatagsRecord _metatagsRecord = null;
```
naveenreddykarri: nit: spacing

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Metric.java
```diff
@@ -86,6 +87,7 @@ public Metric(Metric metric) {
 		setDisplayName(metric.getDisplayName());
 		setUnits(metric.getUnits());
 		setQuery(metric.getQuery());
+                setMetatagsRecord(metric.getMetatagsRecord());
```
naveenreddykarri: nit: spacing before start of line

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Metric.java
naveenreddykarri: nit: spacing before start of line

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -164,9 +181,19 @@ public void put(List<Metric> metrics) {
 			if(!found) {
 				scopesAndMetricsNamesToPut.add(Pair.of(scopeName, metricName));
 			}
+
+                        // Check if metatags are present and populate accordingly
+                        MetatagsRecord mtags = metric.getMetatagsRecord();
+                        if(mtags != null) {
+                            key = mtags.getKey();
+                            found = bloomFilterMetatags.mightContain(key);
```
naveenreddykarri: Do we need to check key != null

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Metric.java
nkunal: I think GitHub is messed up, if you view the raw file, it does not show the spacing. Neither does by editor. https://raw.githubusercontent.com/salesforce/Argus/621d45fd9d9627e64058e16dfaa145c67c4eab77/ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Metric.java

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Metric.java
nkunal: same issue here, I don't see the spacing issue on the editor, but it is visible when doing github review. But its not visible on GitHub when you view the raw file for this PR.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetatagsRecord.java
```diff
@@ -0,0 +1,95 @@
...
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.HashMap;
+import java.util.Map.Entry;
+import java.util.TreeMap;
+import com.salesforce.dva.argus.entity.TSDBEntity.ReservedField;
+
+
+public class MetatagsRecord {
```
dilipdevaraj-sfdc: Can you pls add javadoc for public class

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetatagsRecord.java
```diff
@@ -0,0 +1,95 @@
...
+
+            if (!ReservedField.isReservedField(key)) {
+                result.put(key, entry.getValue());
+            }
+        }
+        return Collections.unmodifiableMap(result);
+    }
+
+    /**
+     * Replaces the metatags for a metric. Metatags cannot use any of the reserved tag names.
```
dilipdevaraj-sfdc: What if the user passes in a reserved field say "units" or "displayname", which Argus can use internally in future ?
We can publish the list of reserved fields to users as documentation, but we might still need to check for reserved fields and not add it to this metaTags map.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/MetatagsSchemaRecordList.java
```diff
@@ -0,0 +1,137 @@
...
+            if(JsonNodeType.ARRAY.equals(hits.getNodeType())) {
+                records = new ArrayList<>(hits.size());
+                Iterator<JsonNode> iter = hits.elements();
+                while(iter.hasNext()) {
+                    JsonNode hit = iter.next();
+                    JsonNode source = hit.get("_source");
+
+                    JsonNode metricMetatagsNodes = source.get(RecordType.METATAGS.getName());
+
+                    // TBD  fix creation of metatags records from json
```
dilipdevaraj-sfdc: Why is this still TBD?

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetatagsRecord.java
dilipdevaraj-sfdc: Ok. I see this is not end user exposed fields , instead it is something an Argus developer controls - ie extracting x no. fields from metric schema - like A,B,C - this should be a well defined finite limit.

nkunal: yup

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/MetatagsSchemaRecordList.java
nkunal: This will be taken care of when do the querying layer. The deserialization code will be invoked at that time. I just put a note there to code it when the querying layer is added.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/MetatagsSchemaRecordList.java
```diff
@@ -0,0 +1,137 @@
...
+                String fieldsData = mapper.writeValueAsString(entry.getValue().getMetatags());
+                SchemaRecordList.addCreateJson(jgen, entry.getKey(), fieldsData);
+            }
+        }
+    }
+
+    static class UpdateSerializer extends JsonSerializer<MetatagsSchemaRecordList> {
+
+        @Override
+        public void serialize(MetatagsSchemaRecordList list, JsonGenerator jgen, SerializerProvider provider)
```
dilipdevaraj-sfdc: metaTags is a map whose key is written as field names into ES
We should add a Javadoc that the metaTags key should be a well defined finite limit say 50 (otherwise this can be misused in open source)

nkunal: The length should not matter, because we take a hash of it and use as the "_id" field when we insert in ES.

dilipdevaraj-sfdc: If the fields for the document are only   _id,mts and cts,  how do we get back the metatag "A","B","C" when we do a query against  this index ?

nkunal: No the fields for the ES document are as follows:
{ "create" : {"_id" : "faae0a8d00d7b6d18f11bd09fd9e09a1"}}
{"meta1":"tagv2","meta2":"tagv1","meta3":"tagv3","cts":15362353,"mts":1536235382637}
Checkout MetatagsSchemaRecordList.java:87 , it puts the meta fields in there.

PR #912, opened 2018-09-12
Made changes to commit to database for every merge operation. Also, reduced the notification cache refresh interval

sundeepsf:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/jpa/DefaultJPAService.java
```diff
@@ -88,7 +88,9 @@ protected DefaultJPAService(AuditService auditService, SystemConfiguration confi
 	public <E extends Identifiable> E mergeEntity(EntityManager em, E entity) {
 		requireArgument(em != null, "The entity manager cannot be null.");
 		requireArgument(entity != null, "The entity cannot be null.");
-		return em.merge(entity);
+		E ret = em.merge(entity);
+		em.flush();
```
dilipdevaraj-sfdc: Are we doing a flush following a merge for required DB calls in other places as well to ensure database is consistent with jpa entity ?

sundeepsf: Yes, I did a search on em.merge and all other places we are doing a flush

PR #917, opened 2018-09-17
W-5349757: remove retention discovery from tags before sending to TSDB

kow1011b:


nkunal:
@kow1011b lets sync up today to discuss this.

PR #919, opened 2018-09-18
 Bug fixes in clear notification flow for GOC and GUS notifications.

sundeepsf:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DataLagMonitor.java
```diff
@@ -51,9 +51,15 @@ public DataLagMonitor(SystemConfiguration sysConfig, MetricService metricService
...
 	public void run() {
 		_logger.info("Data lag monitor thread started");
+		boolean firstTime = true;
 		while (!isInterrupted()) {
 			try {
-				sleep(SLEEP_INTERVAL_MILLIS);
+				if(!firstTime) {
+				    sleep(SLEEP_INTERVAL_MILLIS);
+				}else {
+					sleep(5*1000);
```
dilipdevaraj-sfdc: Can we add comment why we are sleeping for 5sec

dilipdevaraj-sfdc: Please define this as Constant instead of magic number

sundeepsf: I am just waiting a bit for everything else to be initialized

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schedule/DistributedDatabaseSchedulingService.java
```diff
@@ -486,8 +488,8 @@ public void run() {
 					}
 
 					if(!alertEvaluated) {
-						_logger.warn("Notification sent metric is not found in tsdb, so publishing 5 minutes as alert evaluation kpi");
-						publishKPIMetric(fiveMinuteStartTime, new Double(5*60*1000));
+						_logger.warn("Notification sent metric is not found in tsdb, so publishing 4 minutes as alert evaluation kpi");
+						publishKPIMetric(fiveMinuteStartTime, new Double(4*60*1000));
```
dilipdevaraj-sfdc: should this variable be renamed now to fourMinuteStartTime in all occurrences?

dilipdevaraj-sfdc: Why are we changing it from 5 to 4 mins ?

sundeepsf: Because I added a sleep of 1 minute for the alert cache to be refreshed. This thread runs every 5 minutes and so the remaining time is only 4 minutes

PR #921, opened 2018-09-19
 Add handling for no-data scenario for ag-table and some other fixes

prestonfff:
- Add handling for no-data scenario
- Fix pagination info displaying issue


gauravk48:
LGTG

PR #922, opened 2018-09-20
[W-5386828] adding metrics to help investigation 

taozhangsfdc:
https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000005Wd8RIAS/view

Add more metrics to help investigate why alert.scheduled != alert.evaluated when alert.skipped show as zero

The explanation is in the code, and I made small modification to readme.
@sundeepsf

### DIFF
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/MonitorService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/MonitorService.java
@@ -194,6 +194,8 @@ public interface MonitorService extends Service {
         ALERTS_FAILED("argus.core", "alerts.failed"),
         ALERTS_EVALUATION_LATENCY("argus.core", "alerts.evaluation.latency"),
         ALERTS_SKIPPED("argus.core", "alerts.skipped"),
+        ALERTS_PICKEDUP_LIST_SIZE("argus.core", "alerts.pickedupList.size"),
+        ALERTS_PICKEDUP("argus.core", "alerts.pickedup"),
         NOTIFICATIONS_SENT("argus.core", "notifications.sent"),
         TRIGGERS_VIOLATED("argus.core", "triggers.violated"),
         ALERTS_MAX("argus.core", "alerts.max"),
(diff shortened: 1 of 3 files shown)


ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
```diff
@@ -402,6 +413,28 @@ public void updateNotificationsActiveStatusAndCooldown(List<Notification> notifi
 			long jobStartTime = System.currentTimeMillis();
 			long jobEndTime = 0;
 			Long alertEnqueueTimestamp = 0L;
+			
+			/**
```
sundeepsf: this is good to understand for reviewer, not sure if we need to commit such extensive comments for this issue. Maybe a one-liner saying this is to get more data for alert tracking should be enough

ArgusCore/src/main/java/com/salesforce/dva/argus/service/MonitorService.java
```diff
@@ -194,6 +194,8 @@
         ALERTS_FAILED("argus.core", "alerts.failed"),
         ALERTS_EVALUATION_LATENCY("argus.core", "alerts.evaluation.latency"),
         ALERTS_SKIPPED("argus.core", "alerts.skipped"),
+        ALERTS_PICKEDUP_LIST_SIZE("argus.core", "alerts.pickedupList.size"),
+        ALERTS_PICKEDUP("argus.core", "alerts.pickedup"),
```
sundeepsf: Maybe a better name would be "alerts.readFromQueue"

ArgusCore/src/main/java/com/salesforce/dva/argus/service/MonitorService.java
```diff
@@ -194,6 +194,8 @@
         ALERTS_FAILED("argus.core", "alerts.failed"),
         ALERTS_EVALUATION_LATENCY("argus.core", "alerts.evaluation.latency"),
         ALERTS_SKIPPED("argus.core", "alerts.skipped"),
+        ALERTS_PICKEDUP_LIST_SIZE("argus.core", "alerts.pickedupList.size"),
```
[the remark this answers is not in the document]
sundeepsf: We don't need 2 metrics here. You can publish hostname and username as separate tags for a single counter. That will get published as such and we can aggregate on individual dimensions using query

PR #923, opened 2018-09-20
Provide custom discovery retention period (part 1)

kow1011b:
tests will follow.


ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetricSchemaRecord.java
```diff
@@ -42,6 +42,9 @@
  * @author  Tom Valine (tvaline@salesforce.com)
  */
 public class MetricSchemaRecord {
+    public static final String RETENTION_DISCOVERY = "_retention_discovery_";
+    public static final int DEFAULT_RETENTION_DISCOVERY = 45;
```
nkunal: maybe append _DAYS

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetricSchemaRecord.java
```diff
@@ -42,6 +42,9 @@
...
+    public static final int MAX_RETENTION_DISCOVERY = 120;
```
nkunal: maybe append _DAYS

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetricSchemaRecord.java
```diff
@@ -52,6 +55,8 @@
     private String tagKey;
     @JsonProperty("tagv")
     private String tagValue;
+    @JsonProperty(RETENTION_DISCOVERY)
+    private Integer retentionDiscovery;
```
nkunal: Just to be extra careful, could you check if having this as a class var/json prop, will not impact any other places where this class is used.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetricSchemaRecord.java
```diff
@@ -177,6 +197,20 @@ public void setTagValue(String tagValue) {
...
     }
 
+    public Integer getRetentionDiscovery() {
+        return retentionDiscovery;
+    }
+
+    public void setRetentionDiscovery(Integer retentionDiscovery) {
+        if (retentionDiscovery!=null
+                && (retentionDiscovery < DEFAULT_RETENTION_DISCOVERY || retentionDiscovery > MAX_RETENTION_DISCOVERY)) {
+            this.retentionDiscovery = DEFAULT_RETENTION_DISCOVERY;
```
nkunal: If customers send retention value that is less than 45 days, we should use that value instead of 45 days (default value). This is to let customers have a way to expire earlier than the default 45 days. For cases where they deliberately send test metrics and then want to delete them from discovery after a week.
The check we should have is , to cap the retention period to the max_retention

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -390,9 +390,19 @@ protected void implementationSpecificPut(List<Metric> metrics,
...
 			}
 
+			String retention  = metric.getMetatagsRecord()==null?null:metric.getMetatagsRecord().getMetatagValue(MetricSchemaRecord.RETENTION_DISCOVERY);
+			Integer retentionInt = null;
+			if (retention != null) {
+				try {
+					retentionInt = Integer.parseInt(retention);
+				}
+				catch(NumberFormatException e) {
+					_logger.warn("expect _retention_discovery_ to be a numeric value; {} is invalid", retention);
```
nkunal: this could flood the logs and bring argus down, by one bad actor sending invalid retention value, I would change this to debug

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -390,9 +390,19 @@ protected void implementationSpecificPut(List<Metric> metrics,
...
+				}
+			}
 			for(Map.Entry<String, String> entry : metric.getTags().entrySet()) {
 				records.add(new MetricSchemaRecord(metric.getNamespace(), metric.getScope(), metric.getMetric(),
-						entry.getKey(), entry.getValue()));
+						entry.getKey(), entry.getValue(), retentionInt));
 				if(records.size() == _bulkIndexingSize) {
```
nkunal: nitpick: we can do fix the formatting here, and move each param to its own line, will make PR review and code readability a bit easier

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -458,6 +468,8 @@ protected void implementationSpecificPut(List<Metric> metrics,
 
 		List<MetatagsRecord> records = new ArrayList<>(_bulkIndexingSize);
                 for(Map.Entry<String, MetatagsRecord> entry : metatagsToPut.entrySet()) {
+                	//W-5349757: remove this special metatag to prevent it from going to ES
```
nkunal: work item not needed here.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/SchemaService.java
```diff
@@ -236,7 +236,9 @@ static MetricSchemaRecord constructMetricSchemaRecordForType(String str, RecordT
         /** Match against the tag value field. */
         TAGV("tagv"),
         /** Match against the metatags value field. */
-        METATAGS("metatags");
+        METATAGS("metatags"),
+        /** Match against the retention discovery field */
+        RETENTIONDISCOVERY(MetricSchemaRecord.RETENTION_DISCOVERY);
```
nkunal: we could probably keep the name formatting same all across. so RETENTION_DISCOVERY(MetricSchemaRecord.RETENTION_DISCOVERY) instead of RETENTIONDISCOVERY(MetricSchemaRecord.RETENTION_DISCOVERY)

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/MetricSchemaRecordList.java
```diff
@@ -133,12 +133,14 @@ public MetricSchemaRecordList deserialize(JsonParser jp, DeserializationContext
...
 					JsonNode tagkNode = source.get(RecordType.TAGK.getName());
 					JsonNode tagvNode = source.get(RecordType.TAGV.getName());
+					JsonNode retentionNode = source.get(RecordType.RETENTIONDISCOVERY.getName());
 					
 					records.add(new MetricSchemaRecord(namespaceNode == null ? null : namespaceNode.asText(), 
-													   scopeNode.asText(), 
-													   metricNode.asText(), 
-													   tagkNode == null ? null : tagkNode.asText(), 
-													   tagvNode == null ? null : tagvNode.asText()));
+													   	scopeNode.asText(),
```
nkunal: did whitespace formatting got changed here ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/tsdb/AbstractTSDBService.java
```diff
@@ -350,6 +352,12 @@ public void putMetrics(List<Metric> metrics) {
 		List<Metric> fracturedList = new ArrayList<>();
 
 		for (Metric metric : metrics) {
+			MetatagsRecord metatagsRecord = metric.getMetatagsRecord();
+			if (metatagsRecord != null) {
+				//W-5349757: remove this special metatag to prevent it from going to TSDB
```
nkunal: no work item here

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetricSchemaRecord.java
kow1011b: Sure, I will cap it to the max_retention.  What's the expected behavior when customers set it to 0 or negative value?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/MetricSchemaRecordList.java
kow1011b: yeah, besides the addition of retentionNode line, the others are just whitespace difference.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetricSchemaRecord.java
kow1011b: anything < 1 is ignored.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/MetricSchemaRecord.java
kow1011b: I have added serialization/deserialization tests to make sure a similar ObjectMapper as ours can take care of existing records that don't contain the new property.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -1488,6 +1500,7 @@ private ObjectNode _createMappingsNode() {
 		propertiesNode.put(RecordType.TAGK.getName(), _createFieldNode(FIELD_TYPE_TEXT));
 		propertiesNode.put(RecordType.TAGV.getName(), _createFieldNode(FIELD_TYPE_TEXT));
 		propertiesNode.put(RecordType.NAMESPACE.getName(), _createFieldNode(FIELD_TYPE_TEXT));
+		propertiesNode.put(RecordType.RETENTION_DISCOVERY.getName(), _createFieldNode(FIELD_TYPE_TEXT));
```
nkunal: The field_type should be integer/number here, because the nightly job will be doing a math calculation on it while purging stale tags.

ArgusCore/src/test/java/com/salesforce/dva/argus/entity/MetricSchemaRecordTest.java
```diff
@@ -0,0 +1,35 @@
+package com.salesforce.dva.argus.entity;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
nkunal: I would add specific imports instead of wildcard.

ArgusCore/src/test/java/com/salesforce/dva/argus/entity/MetricSchemaRecordTest.java
```diff
@@ -0,0 +1,35 @@
...
+        original.setRetentionDiscovery(50);
+        String s2 = objectMapper.writeValueAsString(original);
+        assertTrue("expect the new json to contain the retention value", s2.contains(MetricSchemaRecord.RETENTION_DISCOVERY));
+
+        MetricSchemaRecord deserialized2 = objectMapper.readValue(s2, MetricSchemaRecord.class);
+        assertEquals("expect the new deserialized object to be equal to the original", original, deserialized2);
+        assertEquals("expect the new deserialized object to have retention value", deserialized2.getRetentionDiscovery().intValue(), 50);
+    }
+
+}
```
nkunal: Fantastic !!

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
kow1011b: I am going to introduce a new type called FIELD_TYPE_INTEGER.  My original concern was the lack of visibility into how these types are being used, and more specifically, whether things would break when they encounter my new type.

ArgusCore/src/test/java/com/salesforce/dva/argus/entity/MetricSchemaRecordTest.java
kow1011b: totally agree.  I have fixed my IDE setting.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
nkunal: new field type sounds good. As you might have guessed, these field types once defined in the index, are then not changeable. Also ignore the foll. if I have already said it:
For existing index, it is important that the new field is added to the settings of that index (out of band), If we don't do that, and just create/update a doc with the new field, then ES will guess its type and we will be stuck with it.

kow1011b:
changes merged in another PR.

PR #931, opened 2018-09-28
Made changes to pass in the QueryContext object which contains details on the underlying tsdb query to all the transforms

sundeepsf:


sundeepsf:
@studanshu , please take a look at this. I made changes to be able to parse the query expression as a tree which is modeled by QueryContext object. Using this we can easily deduce things like start/end times, datacenter for the alert expression etc., without having to rely on regex matches

PR #935, opened 2018-10-02
Setting userdefined2 as prodoutage@salesforce.com if sractionable is …

studanshu:
…checked


ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/notifier/GOCNotifier.java
```diff
@@ -151,6 +151,9 @@ public void sendMessage(Severity severity, String className, String elementName,
 					builder.withActive(true).withCreatedAt(lastNotified);
 				}
 				builder.withLastNotifiedAt(lastNotified);
+				if (srActionable == true) {
+					builder.withUserdefined2("prodoutage@salesforce.com");
```
sundeepsf: can we put this email address in notifier.properties?

PR #942, opened 2018-10-05
discovery service to relay 400 errors from underlying services

kow1011b:
also refactored original method to make it testable.


ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/ElasticSearchSchemaService.java
```diff
@@ -1366,52 +1366,48 @@ protected void enableScopeMetricNamesIndex()
...
-				} catch (IOException ex) {
-					_logger.warn("Failed to close entity stream.", ex);
-				}
+			catch (IOException ex) {
+				throw new SystemException(ex);
 			}
 		}
+
+		//if the response is in the 400 range, use IllegalArgumentException, which currently translates to a 400 error
+		if (statusCode>= HttpStatus.SC_BAD_REQUEST && statusCode < HttpStatus.SC_INTERNAL_SERVER_ERROR) {
```
nkunal: SC_INTERNAL_SERVER_ERROR is 500, we could get 503, or 507, we should adjust for that.

PR #948, opened 2018-10-10
write bloom state and use that to pre-populate bloom on a fresh restart

nkunal:
@naveenreddykarri @colbyguan @dilipdevaraj-sfdc

### DIFF
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
@@ -14,6 +14,12 @@ import java.util.Random;
 import java.util.Set;
 import java.util.HashMap;
 import java.util.Map;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.File;
 
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -48,6 +54,7 @@ public abstract class AbstractSchemaService extends DefaultService implements Sc
 	private static final int DAY_IN_SECONDS = 24 * 60 * 60;
 	private static final int HOUR_IN_SECONDS = 60 * 60;
 
+
 	/* Have three separate bloom filters one for metrics schema, one only for scope names schema and one only for scope name and metric name schema.
 	 * Since scopes will continue to repeat more often on subsequent kafka batch reads, we can easily check this from the  bloom filter for scopes only.
 	 * Hence we can avoid the extra call to populate scopenames index on ES in subsequent Kafka reads.
@@ -74,11 +81,18 @@ public abstract class AbstractSchemaService extends DefaultService implements Sc
 	protected final boolean _syncPut;
 	private int bloomFilterFlushHourToStartAt;
 	private ScheduledExecutorService scheduledExecutorService;
+        private String bfTagsStateFilename;
+
 
 	protected AbstractSchemaService(SystemConfiguration config, MonitorService monitorService) {
 		super(config);
 
 		_monitorService = monitorService;
+
+                String bfStateBaseDir = config.getValue(Property.BF_STATE_BASE_DIR.getName(),
+                                                        Property.BF_STATE_BASE_DIR.getDefaultValue());
+                bfTagsStateFilename = bfStateBaseDir + "/bloomfilter_tags.state." +
+                    config.getValue(SystemConfiguration.ARGUS_INSTANCE_ID, "noid");
 
 		bloomFilterExpectedNumberInsertions = Integer.parseInt(config.getValue(Property.BLOOMFILTER_EXPECTED_NUMBER_INSERTIONS.getName(),
 				Property.BLOOMFILTER_EXPECTED_NUMBER_INSERTIONS.getDefaultValue()));
@@ -101,8 +115,7 @@ public abstract class AbstractSchemaService extends DefaultService implements Sc
 		bloomFilterMetatagsErrorRate =
                     Double.parseDouble(config.getValue(Property.BLOOMFILTER_METATAGS_ERROR_RATE.getName(),
                                                        Property.BLOOMFILTER_METATAGS_ERROR_RATE.getDefaultValue()));
-
-		bloomFilter = BloomFilter.create(Funnels.stringFunnel(Charset.defaultCharset()), bloomFilterExpectedNumberInsertions , bloomFilterErrorRate);
+                createOrReadBloomFilter();
 
 		bloomFilterScopeOnly = BloomFilter.create(Funnels.stringFunnel(Charset.defaultCharset()), bloomFilterScopeOnlyExpectedNumberInsertions , bloomFilterScopeOnlyErrorRate);
 		bloomFilterScopeAndMetricOnly = BloomFilter.create(Funnels.stringFunnel(Charset.defaultCharset()),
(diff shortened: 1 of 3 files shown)


ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -312,6 +325,42 @@ private int getBloomFilterFlushHourToStartAt() {
 		return bloomFilterFlushHourToStartAt;
 	}
 
+    private void createOrReadBloomFilter() {
```
dilipdevaraj-sfdc: Have we tested this to see what is the performance improvement ?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -457,6 +507,9 @@ public void run() {
 
 		private void _flushBloomFilter() {
 			_logger.info("Flushing out bloom filter entries");
+                        // Write the main tags bloom filter to file first before flushing
+                        writeTagsBloomFilterToFile();
```
dilipdevaraj-sfdc: We are only writing state of bloom filter when we do a periodic daily flush via _flushBloomFilter

What about case when we restart schema client before a day has elapsed - the BF will have a state, and we may not be saving this before a restart?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -312,6 +325,42 @@ private int getBloomFilterFlushHourToStartAt() {
...
+            _logger.info("State file for BloomFilter tags NOT present, starting fresh bloom");
+            this.bloomFilter = BloomFilter.create(Funnels.stringFunnel(Charset.defaultCharset()),
+                                                  bloomFilterExpectedNumberInsertions ,
+                                                  bloomFilterErrorRate);
+            return;
+        }
+
+        _logger.info("State file for BloomFilter tags exists, using it to pre-populate bloom");
+        try (InputStream inputStream = new FileInputStream(bfFile)) {
+            this.bloomFilter = BloomFilter.readFrom(inputStream,
```
dilipdevaraj-sfdc: See https://github.com/salesforce/Argus/blob/21f6a391ce5d520c190791b431dc9f8d648c148b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java#L292

In state1
we are adding randomNumber1 to BF key
But in state 2
we are refreshing to randomNumber2 as key in BF
But when we read back state1 BF keys, these will no longer be useful since the keys are supposed to have randomNumber2.
Thus the total number of BF keys will continue to grow linearly, and can cause problems.

dilipdevaraj-sfdc: Even if we to remove randomNumber, when we are reading from previous state of BF instead of having a clean slate are we not introducing a higher probability of false positives ?
If we have schema1 and schema2 hashing to same BF key, and if schema1 were to have appeared first, then schema2 may never be written to ES. (this is why we had random number in first place)
So by continuing to reload BF with previous states schema2 will always appear as false positive since it will be hashing to same key as schema1

Note - we may not be able to entirely rely on running multiple schema committers on different machines to offset false positives.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
nkunal: yes in the first iteration we will do a flush once a day. In the next iteration we will do it 4-6 times per day.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
nkunal: yes good catch ! I will figure out a solution

nkunal: we will keep the randomNumber (will find a solution for preserving the random number, stay tuned). 
So after this PR it won't be any different than existing implementation, since the probability error rate won't be any different than what it was prior to a clean start of bloom

nkunal: @dilipdevaraj-sfdc instead of randomNumber I switched to the hash of the hostname. That way we get uniqueness across multiple hosts, but for given host, we will restart with the same random value.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -101,8 +119,7 @@ protected AbstractSchemaService(SystemConfiguration config, MonitorService monit
 		bloomFilterMetatagsErrorRate =
                     Double.parseDouble(config.getValue(Property.BLOOMFILTER_METATAGS_ERROR_RATE.getName(),
                                                        Property.BLOOMFILTER_METATAGS_ERROR_RATE.getDefaultValue()));
-
-		bloomFilter = BloomFilter.create(Funnels.stringFunnel(Charset.defaultCharset()), bloomFilterExpectedNumberInsertions , bloomFilterErrorRate);
+                createOrReadBloomFilter();
 
 		bloomFilterScopeOnly = BloomFilter.create(Funnels.stringFunnel(Charset.defaultCharset()), bloomFilterScopeOnlyExpectedNumberInsertions , bloomFilterScopeOnlyErrorRate);
```
naveenreddykarri: is the plan we will update all bloomfilters ultimately. Am curious on sizes of generated files for each.

nkunal: the metadata_index bloom filter is the biggest one that will gain from this. The other bloom we won't see much advantage. But the infra code is there to support it if needed.
The size of the main bloomfilter depends on how many entries it has. Bloom grabs chunks as needed and when it fills up it grab further chunks until it reaches the bloom filter size.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
dilipdevaraj-sfdc: HBase and ElasticSearch functionally works the same irrespective if we deploy it as a single instance or multi-instance.
Since Argus is open source, in terms of design this may not be so ideal. If we were to run schema client on only 1 machine, then it will result in lot more false positives since we reuse the hostname, throughout the life-cycle of the process.

dilipdevaraj-sfdc: HBase and ElasticSearch functionally works the same irrespective if we deploy it as a single instance or multi-instance.
Since Argus is open source, in terms of design this may not be so ideal. If we were to run schema client on only 1 machine, then it will result in lot more false positives since we reuse the hostname, throughout the life-cycle of the process.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -312,6 +329,42 @@ private int getBloomFilterFlushHourToStartAt() {
...
+        } catch (IOException io) {
+            _logger.error("tags bloomfilter read error, not using prev state", io);
+            this.bloomFilter = BloomFilter.create(Funnels.stringFunnel(Charset.defaultCharset()),
+                                                  bloomFilterExpectedNumberInsertions ,
+                                                  bloomFilterErrorRate);
+        }
+    }
+
+    private void writeTagsBloomFilterToFile() {
+        File bfTagsFile = new File(this.bfTagsStateFilename);
```
dilipdevaraj-sfdc: Do we want to continue to write the state of BloomFilter to a single file ?
If so, when we retire stale metadata greater than X days from ES via jobs , we do not do that with the BF singular file.
So over a period of time this file will again continue to grow linearly as time elapses. Thus the reload might become slower over time.

nkunal: on each refresh (once every 24 hours) new bloom is created in memory, this new bloom state is then stored in the file, after 24 hours, and it is stored by overwriting the previous file. So we are erasing the previous state (file) each day.

dilipdevaraj-sfdc: It is more to do with BF in memory state, than the BF file itself.
Consider in beginning we have configured BF with max 10000 insertions.
1st day we see 5000 insertions. Next day we will add only new records, and are reloading from file via createOrReadBloomFilter. We never remove older records from BF in memory that are stale. Thus BF in memory will increment second day to say 5500. Third day say 5600....... Here the total BF in memory records and thus size will continue to grow slowly over time.

This is because although we may be creating a new BF object, "createOrReadBloomFilter" reloads the previous days state and thus when we overwrite to file the stale state will still be carried over.

Does this make sense?

nkunal: on the second day, right before flush BF in memory will have 5500 records, those will get written to file. then we BF in memory will have 0 records, it will get populated for the next 24 hours, then these fresh 'N' records will be written to file. So every 24 hours BF in memory as well as BF in file will get fresh records (no carryover) The only thing is BF in file will be lagging by 24 hours.

dilipdevaraj-sfdc: I see that in flush we are doing only writeTagsBloomFilterToFile() and not createOrReadBloomFilter(), so we should be fine like you said in above comment.
However the above problem is still inherent when we do a restart of the process since we call "createOrReadBloomFilter()".
So at the 50th day when the process it restarted we will have all records in BF when we call createOrReadBloomFilter() and some will be stale, since we do not purge this.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
dilipdevaraj-sfdc: probablity of false positive even on single machine should be low since we don't exceed maximum num of records configured for BF.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
dilipdevaraj-sfdc: The file is overwritten in binary encoded format so it will not grow and BF will only have updated data, and not previous state data, so we  are good.

PR #952, opened 2018-10-12
Fixes to address tsdb batch update failures

naveenreddykarri:
also catch all exceptions so that retries work. Fix bug in annotation retries.
Also added some alert code refactoring.

### DIFF
--- a/ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
+++ b/ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
@@ -381,14 +381,14 @@ public class DefaultAlertService extends DefaultJPAService implements AlertServi
 				String logMessage = MessageFormat.format("Failed to deserialize alert {0}. Full stack trace of exception {1}", serializedAlert, ExceptionUtils.getFullStackTrace(e));
 				_logger.warn(logMessage);
 
-				LogAlertStatsOnFailure(DEFAULTALERTID, DEFAULTUSER);
+				logAlertStatsOnFailure(DEFAULTALERTID, DEFAULTUSER);
 
 				continue;
 			}
 
 			if(!_shouldEvaluateAlert(alert, alert.getId())) {
 
-				LogAlertStatsOnFailure(alert.getId(), alert.getOwner().getUserName());
+				logAlertStatsOnFailure(alert.getId(), alert.getOwner().getUserName());
 				continue;
 			}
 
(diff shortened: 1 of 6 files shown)


ArgusCore/src/test/java/com/salesforce/dva/argus/service/CallbackServiceTest.java
```diff
@@ -33,7 +33,7 @@ public void testCallbackNotifier() {
 				"* * * * *");
 		final Trigger trigger = new Trigger(alert, Trigger.TriggerType.GREATER_THAN_OR_EQ, "trigger_name", 2D, 5);
 
-		final String jsonBody = "{ \"uri\" : \"localhost:8080\", \"method\" : \"POST\", \"header\": { \"Content-Type\": \"application/json\" }, \"body\": \"{\\\"triggerName\\\": \\\"«trigger.name>»\\\", \\\"alertName\\\": \\\"«alert.name»\\\"}\",\"template\": \"ST4\"}";
+		final String jsonBody = "{ \"uri\" : \"localhost:9600\", \"method\" : \"POST\", \"header\": { \"Content-Type\": \"application/json\" }, \"body\": \"{\\\"triggerName\\\": \\\"«trigger.name>»\\\", \\\"alertName\\\": \\\"«alert.name»\\\"}\",\"template\": \"ST4\"}";
```
naveenreddykarri: we have a mock http client running on 9600 that echos passed arguments so changing the port so that it can be tested locally. Currently the test does not work, so will make more fixes in a later PR.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
```diff
@@ -382,28 +382,14 @@ public void updateNotificationsActiveStatusAndCooldown(List<Notification> notifi
...
 
-				Map<String, String> tags = new HashMap<>();
-				tags.put("host", HOSTNAME);
-				publishAlertTrackingMetric(Counter.ALERTS_EVALUATED.getMetric(), DEFAULTALERTID, -1.0/*failure*/, tags);
-				tags = new HashMap<>();
-				tags.put(USERTAG, DEFAULTUSER);
-				_monitorService.modifyCounter(Counter.ALERTS_FAILED, 1, tags);
-
-				_monitorService.modifyCounter(Counter.ALERTS_EVALUATED, 1, tags);
+				LogAlertStatsOnFailure(DEFAULTALERTID, DEFAULTUSER);
```
sundeepsf: usually we have function names being with lower case letter

PR #956, opened 2018-10-12
Identifying datalag per dc using expression

studanshu:


ArgusCore/src/main/java/com/salesforce/dva/argus/service/MonitorService.java
```diff
@@ -146,7 +146,7 @@
      *
      * @return isDataLagging boolean flag
      */
-    boolean isDataLagging();
+    boolean isDataLagging(String dataCenter);
```
sundeepsf: should this be a list as some alert expressions can have multiple DCs?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/notifier/GusNotifier.java
```diff
@@ -235,9 +234,7 @@ private void postToGus(Set<String> to, String feed) {
 		}
 	}
 
-	private String generateAccessToken() {
-		// Set up an HTTP client that makes a connection to REST API.
-		HttpClient httpclient = getHttpClient(_config);
+	private static String generateAccessToken(HttpClient httpClient, SystemConfiguration _config) {
```
sundeepsf: what are the changes in GUSNotifer related to?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DataLagMonitor.java
```diff
@@ -20,15 +19,17 @@
...
+	private String _dataLagQueryExpressions;
 
 	private long _dataLagThreshold;
 
 	private String _dataLagNotificationEmailId;
 
 	private String _hostName;
 
-	private boolean isDataLagging = false;
+	private Map<String, Boolean> _isDataLagging = new TreeMap<>();
```
sundeepsf: call it isDataLaggingbyDCMap

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DataLagMonitor.java
```diff
@@ -38,93 +39,112 @@
...
-				_logger.error("Exception thrown in data lag monitor thread - " + ExceptionUtils.getFullStackTrace(e));
-			}
+                    long currTime = System.currentTimeMillis();
+                    List<Metric> metrics = _metricService.getMetrics(_expressionPerDC.get(currentDC), currTime);
+                    boolean isDataLagging = _isDataLagging.get(currentDC);
+                    if (metrics == null || metrics.isEmpty()) {
+                        _logger.info("Data lag detected as metric list is empty");
+                        if (!isDataLagging) {
+                            _isDataLagging.put(currentDC, true);
+                            sendDataLagNotification(currentDC);
```
sundeepsf: Don't send a notification for every datacenter that will be too much. Send notificaiton as a summary only when new DCs are added or existing DCs are removed from the map

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DataLagMonitor.java
```diff
@@ -38,93 +39,112 @@
...
-				}
-				if(isDataLagging) {
-					isDataLagging = false;
-					sendDataLagEmailNotification();
-				}
-			}catch(Exception e) {
-				_logger.error("Exception thrown in data lag monitor thread - " + ExceptionUtils.getFullStackTrace(e));
-			}
+                    long currTime = System.currentTimeMillis();
+                    List<Metric> metrics = _metricService.getMetrics(_expressionPerDC.get(currentDC), currTime);
```
sundeepsf: i think its better to execute these queries using futuretask in parallel, as executing them serially is going to take a while and we spend more time in detecting the data lag. Use something like ExecutorCompletionService

ArgusCore/src/main/java/com/salesforce/dva/argus/service/MonitorService.java
studanshu: @sundeepsf that logic is encoded where I am getting all dc per expression

ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/notifier/GusNotifier.java
studanshu: @sundeepsf if there is a data lag, we are posting on argus chatter page.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DataLagMonitor.java
studanshu: @sundeepsf every dc is initialized with **false**, when the context changes for dc, we sent out a gus notification.

ArgusCore/src/main/java/com/salesforce/dva/argus/util/QueryUtils.java
```diff
@@ -52,4 +54,35 @@ public static QueryContext getQueryContext(String expression, Long relativeTo) {
...
+    public static List<String> getScopesFromExpression(String expression) {
+        return getScopesFromExpression(expression, System.currentTimeMillis());
+    }
+
+    public static List<String> getScopesFromExpression(String expression, long relativeTo) {
+	    return getScopesFromExpression(getQueryContext(expression, relativeTo));
+    }
+
+    private static List<String> getScopesFromExpression(QueryContext queryContext) {
+	    ArrayList<String> scopes = new ArrayList<>();
```
sundeepsf: maybe this can be a set, so that duplicates get filtered out

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DataLagMonitor.java
```diff
@@ -38,93 +41,125 @@
...
+                    sleep(SLEEP_INTERVAL_MILLIS);
+                } else {
+                    // waiting 5 seconds for everything to initialize
+                    sleep(5 * 1000);
+                    firstTime = false;
+                }
+
+                for (String currentDC : _expressionPerDC.keySet()) {
+                    completionService.submit(() -> {
+                        List<Metric> metrics = _metricService.getMetrics(_expressionPerDC.get(currentDC), currTime);
```
sundeepsf: need a try catch here to return an empty list if the getMetrics call fails for some reason

ArgusCore/src/main/java/com/salesforce/dva/argus/service/monitor/DataLagMonitor.java
sundeepsf: Yes, I understand that. What I meant was, even if 5 DCs are detected as lagging in a cycle, then we can just send a single notification for the cycle saying the following datacenters are lagging. That will be less chatty. Remember that this notification is sent by each alert client. So if 20 DCs are lagging, you are potentially looking at 100 notifications

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/DefaultMetricService.java
```diff
@@ -239,34 +240,55 @@ private long _getDatapointsAcrossMetrics(List<Metric> metrics) {
 		return dataPointsSize;
 	}
 
-	public List<String> getDCFromExpression(String expression) {
+	private List<String> getMatchedDCAgainstRegex(List<String> scopes, String regex) {
```
sundeepsf: can we have test case for this method? Like pass in a list of scopes and see if it returns the right DC

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/DefaultMetricService.java
```diff
@@ -239,34 +240,55 @@ private long _getDatapointsAcrossMetrics(List<Metric> metrics) {
...
-				while (m.find()) DC.add(m.group().substring(1, m.group().length() - 1).toUpperCase());
-
-				// If it matches ":argus., then it's argus metrics emitted by PRD."
-				m = Pattern.compile(patterns.get(1)).matcher(currentQuery.getScope());
-				if(m.find()) DC.add(defaultDC);
+			//Two cases: 1. DC can be extracted from scope (don't call discovery service) 2. DC cannot be identified, call discovery service.
+			List<String> currentDCList = getMatchedDCAgainstRegex(scopes, patterns.get(0));
+			if (currentDCList != null) //Case 1
+				finalDCList = (ArrayList<String>) currentDCList;
+			else { //Case 1
```
sundeepsf: case 2

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/DefaultMetricService.java
studanshu: @sundeepsf we have a test case in AlertUtilsTest.java

ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/resources/DashboardResources.java
```diff
@@ -72,7 +71,7 @@
 
 	//~ Instance fields ******************************************************************************************************************************
 
-	private DashboardService dService = ArgusWebServletListener.getSystem().getServiceFactory().getDashboardService();
+	private DashboardService dService = system.getServiceFactory().getDashboardService();
```
sundeepsf: what is this change about?

studanshu: @sundeepsf Making it consistent with other resources. not related to data lag

PR #957, opened 2018-10-16
clean branch for metatags support and expiration timestamp update

kow1011b:
- metatags support in collections endpoint, 
- update expiration timestamp in elastic search when documents are updated


ArgusCore/src/main/java/com/salesforce/dva/argus/service/schema/AbstractSchemaService.java
```diff
@@ -291,6 +291,11 @@ protected String constructKey(String scope, String metric, String tagk, String t
 			sb.append('\0').append(tagv);
 		}
 
+		//there is use case where users simply want to update the retention without touching rest of a metric
```
nkunal: We should probably not use retention in the key, keys should be based solely on the primary fields. Else there will be duplicate docs (the ones that are already existing, and the ones that are sent with retention)

PR #961, opened 2018-10-17
Jpa bugfix

pratiksha-shah:


pratiksha-shah:
withdrawing since it was an older PR

PR #962, opened 2018-10-17
Making the createdDate fix per Sundeep's request

pratiksha-shah:
Edit the createdDate and modifiedDate


ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
```diff
@@ -259,7 +259,7 @@ public void markAlertForDeletion(Alert alert) {
 		alert.setDeleted(true);
 		alert.setEnabled(false);
 		alert.setName(alert.getName() + System.currentTimeMillis());
-
+		alert.setModifiedDate(new Date());
```
naveenreddykarri: I think we also need to set modifiedDate for dashboard update.

PR #969, opened 2018-10-23
remove Refocus notifier from UI

yjanezou:
@nikhilwahi @gauravk48


gauravk48:
@prestonfff Can you please take a look at this change?

PR #974, opened 2018-10-24
Ensuring proper validation of email while updating the notification

studanshu:


ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Notification.java
```diff
@@ -443,8 +438,16 @@ public void setSubscriptions(List<String> subscriptions) {
...
         for(String currentSubscription: subscriptions) {
             if (this.getNotifierName().equals(AlertService.SupportedNotifier.GUS.getName())) {
-                if (currentSubscription.length() < 10)
-                    throw new IllegalArgumentException("GUS subjectId is incorrect.");
+                if (currentSubscription.isEmpty() || currentSubscription.length() < 10)
+                    throw new IllegalArgumentException("GUS Subscription has to contain subjectId with more than 10 characters.");
+            } else if (this.getNotifierName().equals(AlertService.SupportedNotifier.EMAIL.getName())) {
+                if (currentSubscription.isEmpty() || !currentSubscription.matches(EMAIL_REGEX)) {
+                    String errorMessage = MessageFormat.format("Email Address {0} is not allowed according to Regex {1}.",
+                            currentSubscription, EMAIL_REGEX);
```
nkunal: Please add unit test for setSubscriptions() method by passing is various email address in the subscriptions. Pick a few different emails formats and verify the regex is working and/or the correct exception is thrown

studanshu: @nkunal added.

ArgusCore/src/main/java/com/salesforce/dva/argus/entity/Notification.java
```diff
@@ -227,45 +222,45 @@ public Notification deserialize(JsonParser jp, DeserializationContext ctxt) thro
 
 	// We allow a-zA-Z0-9-_+. in the name, then @ then a-zA-Z0-9- followed by . and a-zA-Z0-9.
 	// ToDo Consider email.contains("@") if we see more issues in future
-	private static final String EMAILREGEX = "[a-zA-Z0-9\\-\\_\\+\\.]+@[a-zA-Z0-9\\-]+\\.[a-zA-Z0-9]+";
+	private static final String EMAIL_REGEX = "[a-zA-Z0-9\\-\\_\\+\\.]+@[a-zA-Z0-9\\-\\.]+\\.[a-zA-Z0-9]+";
```
naveenreddykarri: @nkunal @sundeepsf  Do we be conservative in our checks and just switch to email.contains("@"). It is in general very difficult to validate if a given string is valid email format.

https://en.wikipedia.org/wiki/Email_address#Local-part

Some Examples
.foo@bar, fo..o@bar are invalid, but our Regex will allow it.
We only allow a subset of the special characters that are allowed.
!#$%&'*+-/=?^_`{|}~

ArgusWebServices/src/main/java/com/salesforce/dva/argus/ws/resources/AlertResources.java
```diff
@@ -708,35 +708,39 @@ public AlertDto updateAlert(@Context HttpServletRequest req, @PathParam("alertId
...
-				int index = alert.getNotifications().indexOf(notification);
+					Alert alert = alertService.updateAlert(oldAlert);
+					int index = alert.getNotifications().indexOf(notification);
 
-				return NotificationDto.transformToDto(alert.getNotifications().get(index));
+					return NotificationDto.transformToDto(alert.getNotifications().get(index));
+				}
 			}
+			throw new WebApplicationException("The notification does not exist.", Response.Status.NOT_FOUND);
+		} catch (Exception e) {
```
naveenreddykarri: this change can be skipped till we handle individual errors.

PR #976, opened 2018-10-24
W-5463837 Verify what changes are needed to make Callback Notifications work

naveenreddykarri:


studanshu:
@naveenreddykarri should history object be a data member of NotificationContext classes? As we do in case for alerts and triggers?

PR #977, opened 2018-10-24
Script that helps update all version numbers (POM and RPM) when doing releases

colbyguan:
Couldn't find a way to do multiline-regex with `sed`, and figuring out a Maven way of doing it would have taken much longer, so here is a Python script. It works like this:

image

Intended to work on both Python 2 and 3. Tested with versions:
* `2.7.15`
* `3.6.3`

Checks user inputs for proper format

### DIFF
--- a/update-version
+++ b/update-version
@@ -1,0 +1,126 @@
+#!/usr/bin/env python
+import re
+import sys
+
+class termcolors:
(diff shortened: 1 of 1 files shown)


nkunal:
You can see me smiling .. a lot. thank you

PR #987, opened 2018-10-29
Transform fixes

sundeepsf:
1. Fixed the FILL, FILL_CALCULATE and PROPAGATE transforms to fill the entire query interval, regardless of number of datapoints
    
    2. Fixed the DOWNSAMPLE transform to auto-extract start and end times from the query itself
    
    3. Added a new option to DOWNSAMPLE transform called #abs# which will generate fixed size intervals without rounding up the times to match the hourly/day boundaries


ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/FillTransform.java
```diff
@@ -244,7 +248,7 @@ private long _parseStartAndEndTimestamps(String timeStr, long relativeTo) {
         for (Metric metric : metrics) {
             Metric newMetric = new Metric(metric);
 
-            newMetric.setDatapoints(_fillMetricTransform(metric, windowSizeInSeconds, offsetInSeconds, value));
+            newMetric.setDatapoints(_fillMetricTransform(queryContext, metric, windowSizeInSeconds, offsetInSeconds, value));
```
dilipdevaraj-sfdc: @sundeepsf 
Is there a scope for user abuse with all the transforms that now leverage queryContext to get user specified time like FILL,....
shouldn't we always be doing a defensive programming if time range is large (we can define something for this) then always throw an exception. This way there is no heap exhaustion due to bad query similar to what we do for a vanilla Argus query with no Argus transforms.

sundeepsf: Actually our 2 million datapoint limit will kick in if the query interval is large and will block that query

dilipdevaraj-sfdc: @sundeepsf 
Since we are anyways modifying FILL and other transforms, it would still be good to have explicit check.
If you are to use synthetic data 
eg) FILL(#-300d#,#-0d#,#1s#,#0m#,#100#)
The query just hangs for a very long time on the UI, and we don't throw an exception for long user query.

PR #989, opened 2018-10-30
fix the keyword conflicts in JMX exporter

taozhangsfdc:
Avoid the possibility that user use tag that conflict with the build-in keywords of the exporter
This is tested in shared2-argusalertdev1-1-prd.eng.sfdc.net

@sundeepsf 
@naveenreddykarri


ArgusCore/src/main/java/com/salesforce/dva/argus/service/alert/DefaultAlertService.java
```diff
@@ -737,12 +737,12 @@ public void sendNotification(Trigger trigger, Metric metric, History history, No
 
 		Map<String, String> tags = new HashMap<>();
 		tags.put("status", "active");
-		tags.put("type", SupportedNotifier.fromClassName(notification.getNotifierName()).name());
+		tags.put("notify-target", SupportedNotifier.fromClassName(notification.getNotifierName()).name());
```
naveenreddykarri: since we have to type this in UI, do we just call it target instead of notify-target?

PR #999, opened 2018-11-02
Rate transform implementation

rajsarkapally:


rajsarkapally:
@sundeepsf @dilip-devaraj  can you please review

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/RateTransform.java
```diff
@@ -0,0 +1,164 @@
...
+import com.salesforce.dva.argus.util.QueryContext;
+import com.salesforce.dva.argus.util.QueryUtils;
+/**
+ * It provides methods to implement Rate transform
+ * @author Raj Sarkapally (rsarkapally@salesforce.com)
+ *
+ */
+public class RateTransform implements Transform{
+	private static long DEFAULT_INTERVAL=60*1000;
+	private static boolean DEFAULT_SKIP_NAGATIVE_VALUES=true;
```
sundeepsf: typo NEGATIVE

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/RateTransform.java
```diff
@@ -0,0 +1,164 @@
...
+		}
+	}
+
+	private double getInterpolatedvalue(Entry<Long, Double> prevDP, Entry<Long, Double> nextDP, long timestamp){
+		double slope = (nextDP.getValue()-prevDP.getValue())/(nextDP.getKey()-prevDP.getKey());
+		double result = prevDP.getValue() + slope*(timestamp-prevDP.getKey()); 
+		return result;
+	}
+
+	private static long _getTimeIntervalInMilliSeconds(String interval) {
```
sundeepsf: there is already a method in TransformUtil for this. You can reuse that

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/RateTransform.java
```diff
@@ -0,0 +1,164 @@
...
+		Double prev = null, curr=null;
+		for (Entry<Long, Double> entry : sortedDatapoints.entrySet()) {
+			curr = entry.getValue();
+			if (prev !=null){
+				result.put(entry.getKey(), curr - prev);
+			}
+			prev = curr;
+		}
+		//Add first DP using interpolation
+		if(interpolateFirstDP) {
```
sundeepsf: isn't the first DP already added by addFirstNLastDatapointsIfMissing method?

sundeepsf: I don't think we need to populate the rate for first datapoint. For example if user passes only two data points, we just need to return 1 datapoint for rate. That is how derivative also works.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/RateTransform.java
```diff
@@ -0,0 +1,164 @@
...
+		return result;
+	}
+
+	private TreeMap<Long, Double> performDifferentiation(TreeMap<Long, Double> sortedDatapoints, boolean interpolateFirstDP, long intervalInMilli) {
+		TreeMap<Long, Double> result = new TreeMap<>();
+		Double prev = null, curr=null;
+		for (Entry<Long, Double> entry : sortedDatapoints.entrySet()) {
+			curr = entry.getValue();
+			if (prev !=null){
+				result.put(entry.getKey(), curr - prev);
```
sundeepsf: here are you inherently assuming that interpolation is always set to true? What is that is set to false and there are missing datapoints in between?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/RateTransform.java
rajsarkapally: this is done

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/RateTransform.java
rajsarkapally: This is done

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/RateTransform.java
rajsarkapally: When we are interpolating all data points in between first and last, why not first dp?if first 10 DPs are missing, we anyway interpolate 9 DPs.. if user does not want interpolation, anyways first DP will not appear.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/RateTransform.java
rajsarkapally: Argus rate function should be similar to Prometheus and by default interpolation is applied in prometheus

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/RateTransform.java
rajsarkapally: Let this be little different from derivative so both derivative and rate will not look like same

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/RateTransform.java
sundeepsf: But then why are you accepting a boolean param for doing interpolation or not from the user?

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/RateTransform.java
sundeepsf: Its ok to interporate to get the first datapoint before applying rate, but when you calculate the rate the number of datapoints in your result will be (n-1) where n is the number of datapoints before applying the rate function.

ArgusCore/src/main/java/com/salesforce/dva/argus/service/metric/transform/RateTransform.java
rajsarkapally: these constants are optional.. if rate function is called without any params, it will exactly behave  like prometheus (thats why interpolation is enabled by default). those parameters will give more flexibility.

sundeepsf: i know, if i pass in false for interpolation, then your rate function will return incorrect results as you are not calculating the slope.

rajsarkapally: I do not want to use slope if there is no interpolation.. Here is the reason
Timestamp .   value
1                          1
2                         2
5                          5

If i use slope, the result will be
2        1
5 .      1 . (Using slope, the value at timestamp 4 is 4)

Let this metric is no.of requests handled by web server.
now if user want count total requests, then he would get 1+1 = 2 but webserver processed 4 requests.

anyways, the default behavior is using slope so user can perform rate with slope or without slope. I will provide good documentation for this

It is not possible to cover all use cases, so default implementation is same as prometheus and more flexibility with constants.
So let our customers use this function and we will change the behavior based on their feedback

